PR #8, opened 2016-08-12
correct the case that properties is NULL for clCreateContextFromType

guoyejun:
according to spec, for function clCreateContextFromType, properties
can also be NULL in which case the platform that is selected is
implementation-defined. While the current code returns CL_INVALID_PLATFORM.

### DIFF
--- a/icd.c
+++ b/icd.c
@@ -198,6 +198,12 @@ Done:
 
 void khrIcdContextPropertiesGetPlatform(const cl_context_properties *properties, cl_platform_id *outPlatform)
 {
+    if (properties == NULL && khrIcdVendors != NULL)
+    {
+        *outPlatform = khrIcdVendors[0].platform;
+        return;
+    }
+
     const cl_context_properties *property = (cl_context_properties *)NULL;
     *outPlatform = NULL;
     for (property = properties; property && property[0]; property += 2)
--- a/icd_dispatch.c
+++ b/icd_dispatch.c
@@ -218,11 +218,6 @@ clCreateContextFromType(const cl_context_properties * properties,
     // determine the platform to use from the properties specified
     khrIcdContextPropertiesGetPlatform(properties, &platform);
 
-    if (properties == NULL && khrIcdVendors != NULL)
-    {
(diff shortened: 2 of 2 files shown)


kepatil:
It make more sense to do the NULL check inside khrIcdContextPropertiesGetPlatform instead.

guoyejun:
I did so because khrIcdContextPropertiesGetPlatform is also called inside function clGetGLContextInfoKHR. There is on explicit words at https://www.khronos.org/registry/cl/sdk/1.2/docs/man/xhtml/clGetGLContextInfoKHR.html to say something like 'properties can also be NULL in which case the platform that is selected is implementation-defined', just as clCreateContextFromType does. 

I just checked it again and found 'CL_INVALID_PLATFORM if properties is NULL and no platform could be selected' at spec of clCreateContext. Looks that it gives an implicit meaning which is the same as the explicit words at clCreateContextFromType.  I'll change the code accordingly.

icd.c
```diff
@@ -198,6 +198,12 @@ void khrIcdVendorAdd(const char *libraryName)
 
 void khrIcdContextPropertiesGetPlatform(const cl_context_properties *properties, cl_platform_id *outPlatform)
 {
+    if (properties == NULL && khrIcdVendors != NULL)
```
guoyejun: from spec of clCreateContext, CL_INVALID_PLATFROM is returned for two cases:
- if properties is NULL and on platform could be selected
- if platform value specified in properties is not a valid platform.

it does not touch the case that properties is non-NULL but does not contain a CL_CONTEXT_PLATFORM property. My understanding is to select an implementation-defined platform for such case.

As in this code review, as the subject suggested, it is for the case that properties is NULL, maybe we can firstly merge this change, and then add another change for the case that properties is not NULL.

bashbaug: I'm fine solving the NULL properties case now and solving the non-NULL (but no CL_CONTEXT_PLATFORM) case later, if needed.

guoyejun:
if no other comments, could the patch be accepted? thanks.

PR #14, opened 2017-04-19
Get vendor path from env var OPENCL_VENDOR_PATH, if it exists

jllanfranchi:
Allows behavior as described here: https://wiki.tiker.net/OpenCLHowTo#Per-user_ICD_registry.3F at least for Linux. If the env var `OPENCL_VENDOR_PATH` is not set, then it defaults to the previously hard-coded values.

### DIFF
--- a/icd_linux.c
+++ b/icd_linux.c
@@ -57,11 +57,17 @@ void khrIcdOsVendorsEnumerate(void)
 {
     DIR *dir = NULL;
     struct dirent *dirEntry = NULL;
+    char *vendorPath = NULL;
+
+    vendorPath = getenv("OPENCL_VENDOR_PATH");
+    if  (NULL == vendorPath)
+    {
 #ifdef __ANDROID__
-    char *vendorPath = "/system/vendor/Khronos/OpenCL/vendors/";
+        vendorPath = "/system/vendor/Khronos/OpenCL/vendors/";
 #else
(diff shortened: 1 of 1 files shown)


kepatil:
Appreciate your desire and effort to improve the ICD loader, @jllanfranchi.

A similar patch was discussed some time ago in the OpenCL working group. The consensus was that adding such an environment variable introduces a system-wide backdoor similar to LD_LIBRARY_PATH. It will cause concerns similar to those raised about LD_LIBRARY_PATH, ranging anywhere from being a mild annoyance to being a security risk. The working group does not want to deal with the concerns unless there is a strong justification to do so.

You are probably aware that there are alternative ways to achieve what your patch does, especially as temporary measures for debugging. For instance, if the ICD entry in /etc/OpenCL/vendors contains a relative path then setting LD_LIBRARY_PATH can pick up the OpenCL implementation library from any user path. Users are also free to use your patch (or similar) to build their own ICD loader and pick it up via LD_LIBRARY_PATH.

TL;DR: This patch was duly considered and its potential hassle outweighs its utility. So I am not accepting it into the official sources.

jllanfranchi:
Sorry for the very delayed response, but I figured it's never too late for a thanks: So thank you for your detailed response and explanation of why this isn't considered a good idea to merge.

PR #16, opened 2017-04-23
CMake improvements

chfast:
This PR add option to disable build tests and to build the OpenCL ICD loaded as a static library.
It also fixes crashes in the icd_test_loader when it cannot find the test platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required (VERSION 2.6)
 
 project (OPENCL_ICD_LOADER)
 
+option (TESTS "Enable building tests" ON)
+
 set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set (CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
@@ -27,5 +29,7 @@ endif ()
 
 target_link_libraries (OpenCL ${CMAKE_DL_LIBS})
 
-enable_testing()
-add_subdirectory (test)
+if (TESTS)
+    enable_testing ()
+    add_subdirectory (test)
+endif ()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required (VERSION 2.6)
 
 project (OPENCL_ICD_LOADER)
 
+option (BUILD_SHARED_LIBS "Build shared libs" ON)
 option (TESTS "Enable building tests" ON)
 
 set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
@@ -20,7 +21,7 @@ endif ()
 # OR copy OpenCL headers to ./inc/CL/
 include_directories (./inc)
 
-add_library (OpenCL SHARED ${OPENCL_ICD_LOADER_SOURCES})
+add_library (OpenCL ${OPENCL_ICD_LOADER_SOURCES})
 set_target_properties (OpenCL PROPERTIES VERSION "1.2" SOVERSION "1")
 
 if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
--- a/test/loader_test/CMakeLists.txt
+++ b/test/loader_test/CMakeLists.txt
@@ -12,4 +12,5 @@ add_executable (icd_loader_test
     test_clgl.c
     test_image_objects.c )
 
-target_link_libraries (icd_loader_test OpenCL IcdLog)
+find_package(Threads)
+target_link_libraries (icd_loader_test OpenCL IcdLog ${CMAKE_THREAD_LIBS_INIT})
--- a/test/loader_test/test_create_calls.c
+++ b/test/loader_test/test_create_calls.c
@@ -174,10 +174,15 @@ int test_clGetPlatformIDs(const struct clGetPlatformIDs_st* data)
                 &param_val_ret_size );  
 
         if (ret_val == CL_SUCCESS ){
-            if(!strcmp(platform_name, "ICD_LOADER_TEST_OPENCL_STUB")) {
+            if (!strcmp(platform_name, "ICD_LOADER_TEST_OPENCL_STUB")) {
                 platform = all_platforms[i];                
             }
         }
+    }
+
+    if (!platform) {
+        // The stub OpenCL not found.
+        return -2;
     }
 
 #if ENABLE_MISMATCHING_PRINTS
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ cmake_minimum_required (VERSION 2.6)
 
 project (OPENCL_ICD_LOADER)
 
+option (BUILD_SHARED_LIBS "Build shared libs" ON)
```
bashbaug: I don't see where this option is used, so it looks like it's only possible to build a static library with these changes.

What is the usage model for the static library?  The nice thing about a dynamic / shared library is that it can be updated asynchronously from applications.  With a static library there's a very real risk that applications may stop working - see the recent HKR changes in this project, as an example.

chfast: The `BUILD_SHARED_LIBS` is the CMake "magic" option to build libraries as shared, OFF by default. So this change make it ON by default to keep the same behavior as before. But it also allows to build this as a static library if someone wants to.

"updated asynchronously"? What do you mean? Windows Update?

kepatil: The ICD loader is primarily designed to be a shared library (OpenCl.dll or libOpenCL.so). It is meant to be shipped either as a part of a vendor driver installation (Windows, Linux) or as part of a platform-specific package that contains the loader (Linux distros). What @bashbaug means is that the shared library can be updated by updating such installations (the display driver on Windows, for example), independent of whether the application is updated or not. IOW the update of the shared library is "asynchronous" w.r.t. to the application update.

@chfast, can you explain your use case where using a statically linked ICD loader is more useful compared to using the shared library version? There are disadvantages to doing that as @bashbaug mentioned above.

chfast: Yes, I'm using this build as a static library in [ethminer](https://github.com/ethereum-mining/ethminer). In the and I have a single binary for every OS. And it will not crash in case OpenCL is not installed in the system.

kepatil: I see that the problem you are trying to solve applies to shared libraries in general, not just the OpenCL ICD loader. Any application that links to a shared library needs the library to be installed on the system or it will fail to launch, which is normal user expectation.

hydra3333: I wish to build ffmpeg as a single STATIC .exe file including the ICD loader.
I can do this with some juggling, .  When it runs it appears to look for the manufacturer installed  OpenCL.DLL and load that.
That is one reason why the ICD loader needs to be built static, baked into in the ffmpeg exe - to avoid dll hell.

chfast: Can you explain me how do you see the OpenCL loader to be shipped to Windows as a shared library? Via Windows Update?

kepatil: @chfast I am assuming you are asking me and not @hydra3333.

The OpenCL loader library (OpenCL.dll) on Windows is installed by OpenCL driver installer from a vendor like Intel or NVIDIA. Windows Update may install or update OpenCL.dll as part of installing or updating the vendor's driver. Windows Update by itself is not the primary mechanism to ship OpenCL.dll.

kepatil: @hydra3333 Can you clarify why dynamically loading OpenCL.dll does not solve the problem for you?

If I got you correctly, with dynamic loading your executable will not crash even when OpenCL.dll is not present. The executable will also benefit from future updates to the ICD loader when the DLL is updated. Sounds like a win-win to me instead of linking statically.

chfast: Those install vendor's implementations of OpenCL, not a generic loader, right?

Anyway, this discussion is quite a waste of time. Are these code changes an issue or not?

kepatil: Vendor installers install two things: The OpenCL ICD loader (OpenCL.dll) and the OpenCL vendor implementation (<vendor-specific-name>.dll). The former is expected to be built using ICD loader source from Khronos (which is this repository, where we are talking right now).

So the OpenCL.dll is generic (vendor-agnostic) even if it comes from a vendor installer. The vendor-specific implementation resides in the second DLL and it gets loaded dynamically by the ICD loader.

@chfast I understand your frustration with the delay in accepting this change but please bear with us for now. The OpenCL working group appreciates your patch and we are actively discussing it within the group.

In this discussion with you and others we are trying to make sure we understand the rationale and use cases for building and using the ICD loader as a static library in shipped applications, while we know there are drawbacks to that approach compared to using a shared library.

kepatil:
@chfast @hydra3333 In addition to this proposed change, do you think it would be helpful to have a document and/or sample code explaining how the ICD Loader is supposed to be distributed and used by applications? The long discussion on the change seems to suggest that it would be useful.

hydra3333:
OK, there's a couple of examples. Here's one:

Over at https://github.com/DeadSix27/python_cross_compile_script this very smart person provides a script which cross-compiles a 64bit version of the ubiquitous ffmpeg video processing program as one large statically linked program with no DLLs (otherwise there'd be heaps and heaps of DLLs).  The khronos icd loader is built as one of the many many modules and statically linked into the single ffmpeg.exe. When ffmpeg is run, the ICD loader module in ffmpeg loads the manufacturer OpemCL.DLL (installed with the driver, eg nvidia's). 

I've also a hacked up copy of another smart person's script https://github.com/rdp/ffmpeg-windows-build-helpers in which I appreciatively re-use deadsix27's approach and add a couple of kludges to make both a 32bit build (which unfortunately relies on an older commit of khronos icd loader) and a 64bit build which uses the current git head.

chfast:
I can give an example how I'm using it (can write it down later on), but I'm not sure this is the example you want.

I'm using Hunter package manager. The "hunterized" version of OpenCL-ICD-Loader contains my CMake changes: https://github.com/hunter-packages/OpenCL-ICD-Loader/pulls?q=is%3Apr+is%3Aclosed.

In the end it is easy as:

```cmake
hunter_add_package(OpenCL)
find_package(OpenCL CONFIG REQUIRED)
target_link_libraries(ethash-cl PRIVATE OpenCL::OpenCL)
```

By default, the Hunter package is built as static lib.

kepatil:
Thanks for the use case examples. They help us understand why static linking may be useful despite the drawbacks.

I am going to accept this patch but with an additional comment explaining the pitfalls of linking to the ICD Loader statically.

chfast:
@kepatil Do you want me to add the comment? If yes, can you write the comment for me?

kepatil:
Sorry, I should have been more clear. I have written one already and circulated for internal review in the working group. Hoping to get it approved quickly.

chfast:
Someone reported that ROCm is not working when this ICD loader is built-in as a static library. We might switch to classic approach as recommended here. But is there a way to build an executable in a way it will not crash in case OpenCL shared library is not available in user's operating system?

kepatil:
> Someone reported that ROCm is not working when this ICD loader is built-in as a static library. We might switch to classic approach as recommended here. But is there a way to build an executable in a way it will not crash in case OpenCL shared library is not available in user's operating system?

Applications can try loading the library at runtime (dlopen or equivalent) to take some graceful action (continue without the library or exit) if the dynamic library is not present on the system.

PR #20, opened 2017-06-19
Add Universal Windows Support to OpenCL

ofirc:
**TL;DR: OpenCL ICD Loader now enumerates HKR registry keys (see below)
in addition to "Khronos\OpenCL\Vendors" registry key.**

Starting from Microsoft Windows 10 RS3/RS4 (Redstone), the Universal INF
does not support AddReg directives to HKLM, and instead require that
drivers nest driver-specific keys under their assigned HKR key.

This patch adds OpenCL platforms enumeration under display adapter HKR
keys, namely under:
HKLM\SYSTEM\ControlSet\Control\Class\{4d36e968-e325-11ce-bfc1-08002be10318}\0000\OpenCLDriverName[Wow]
...
HKLM\SYSTEM\ControlSet\Control\Class\{4d36e968-e325-11ce-bfc1-08002be10318}\000N\OpenCLDriverName[Wow]

Similar to how OpenGL ICDs are detected on Windows by the OpenGL ICD Loader.

This is done *in addition* to the old school "Khronos\OpenCL\Vendors"
enumeration in order to remain backward compatible with older drivers, as
well as other non Graphics Drivers vendors (e.g. Intel OpenCL Experimental
Platform, intelopencl64_2_1.dll).

### DIFF
--- a/icd_windows_hkr.c
+++ b/icd_windows_hkr.c
@@ -41,12 +41,11 @@
 #include <SetupAPI.h>
 #include <devguid.h>
 
-#define KHR_SAFE_RELEASE(mem)         \
-    do                                \
-    {                                 \
-        if (mem)                      \
-            free(mem);                \
-        mem = NULL;                   \
+#define KHR_SAFE_RELEASE(mem)       \
+    do                              \
+    {                               \
+        free(mem);                  \
+        mem = NULL;                 \
     } while (0)
 
 static const char HKR_PREFIX[] = "SYSTEM\\CurrentControlSet\\Control\\Class\\";
--- a/icd_windows_hkr.c
+++ b/icd_windows_hkr.c
@@ -52,7 +52,7 @@ static const char HKR_PREFIX[] = "SYSTEM\\CurrentControlSet\\Control\\Class\\";
 #ifdef _WIN64
 static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverName";
 #else
-static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverNameWow"
+static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverNameWow";
 #endif
 
 // Given a display adapter HKR (GUID\000x), returns the full registry path
--- a/icd_windows_hkr.c
+++ b/icd_windows_hkr.c
@@ -49,7 +49,7 @@
         mem = NULL;                   \
     } while (0)
 
-static const char HKR_PREFIX[] = "SYSTEM\\ControlSet001\\Control\\Class\\";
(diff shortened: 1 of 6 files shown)


bashbaug:
More information about this pull request is here:

https://docs.microsoft.com/en-us/windows-hardware/drivers/install/using-a-universal-inf-file

ofirc:
Updated the HKR branch to use CurrentControlSet instead of ControlSet001.

icd_windows_hkr.c
```diff
@@ -0,0 +1,299 @@
...
+#include "icd.h"
+#include "icd_windows_hkr.h"
+#include <windows.h>
+#include <SetupAPI.h>
+#include <devguid.h>
+
+#define KHR_SAFE_RELEASE(mem)         \
+    do                                \
+    {                                 \
+        if (mem)                      \
```
kepatil: The NULL pointer check is redundant because free() does the check for you.

ofirc:
@kepatil Thanks for the review. I've updated the branch with "8861d54: Remove redundant NULL check from KHR_SAFE_RELEASE".

ofirc:
@kepatil @bashbaug
Hey Guys, any updates on this?

kepatil:
@ofirc Your code changes and the new registry keys are under review in the OpenCL working group. I will update here when the reviews are done.

ofirc:
@kepatil great, thanks a lot!

ofirc:
@bashbaug @kepatil
Guys,
Please hold the integration, OCL and other GFX driver components need to be componentized and have their own Component INF files [1].

What this means is that the Base INF (the Gfx Driver's INF file that gets installed) gets smaller, and now it should reference a new INF file for OCL by using the INF AddComponent directive.

This means that the registry keys introdced by the OCL component will change from HKR keys:
HKLM\SYSTEM\ControlSet\Control\Class{4d36e968-e325-11ce-bfc1-08002be10318}\0000\OpenCLDriverName[Wow]
to SW Component device ID entries:
HKLM\SYSTEM\CurrentControlSet\Control\Class\{5c4c3332-344d-483c-8739-259e934c9cc8}\0000

Currently there is no way to navigate from the former (driver HKR keys) to the latter (SW component device ID) and vice versa. This means that enumerating display adapters is just not enough.

There is an ongoing discussion with MSFT about what should be the right approach, and until this gets resolved and finalized we should hold back the merge.

Thanks,
Ofir

[1] https://docs.microsoft.com/en-us/windows-hardware/drivers/install/adding-software-components-with-an-inf-file

ofirc:
Abandoning this PR in favor of https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/21

ofirc:
Abandoned this PR in favor of https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/21

PR #21, opened 2017-07-21
Added OpenCL Universal Driver Support for Win10 RS3

ofirc:
Starting from Windows 10 RS3+ Graphics Drivers are no longer allowed to
write HKLM entries, and hence the current registry key:
HKLM\SOFTWARE\Khronos\OpenCL\Vendors

Is no longer applicable to graphics drivers for registering their ICD.
Instead, graphics drivers can write to either of the 2 registry keys
below:
1) Display Adapter HKR
2) Software Components HKR

Graphics Drivers (starting from RS3) have dedicated component INF files and are
referenced via the CopyINF directive and/or (optional) "INF AddComponent directive".

Every such component, e.g. OpenCL instance (ICD), is assigned a unique device
instance ID (e.g. 0000, 0001, etc) under the "Software Component" GUID and use it to
store the path to its ICD.

Therefore we enumerate ICDs in the following places (and also in the following order):
1) Display Adapters HKRs (using the display adapter GUID)
2) Software Component HKRs (using the CM_* APIs to traverse the device tree)
3) The usual "Vendors" registry key (see above)

### DIFF
--- a/icd_windows_hkr.c
+++ b/icd_windows_hkr.c
@@ -1,0 +1,369 @@
+/*
+ * Copyright (c) 2017 The Khronos Group Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software source and associated documentation files (the "Materials"),
+ * to deal in the Materials without restriction, including without limitation
+ * the rights to use, copy, modify, compile, merge, publish, distribute,
+ * sublicense, and/or sell copies of the Materials, and to permit persons to
+ * whom the Materials are furnished to do so, subject the following terms and
+ * conditions:
+ *
+ * All modifications to the Materials used to create a binary that is
+ * distributed to third parties shall be provided to Khronos with an
+ * unrestricted license to use for the purposes of implementing bug fixes and
+ * enhancements to the Materials;
+ *
+ * If the binary is used as part of an OpenCL(TM) implementation, whether binary
+ * is distributed together with or separately to that implementation, then
+ * recipient must become an OpenCL Adopter and follow the published OpenCL
+ * conformance process for that implementation, details at:
+ * http://www.khronos.org/conformance/;
+ *
+ * The above copyright notice, the OpenCL trademark license, and this permission
+ * notice shall be included in all copies or substantial portions of the
+ * Materials.
+ *
+ * THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


ofirc:
@bashbaug @kepatil 
This is just a preliminary PR to provide some time ahead for reviewing the code.
The code above was approved by Microsoft for enumerating the ICD entries on Win10 RS3.

There are some opens until we can call this complete:
1) Currently the ICD Loader doesn't link against OneCore.lib and hence is not fully Win10 compliant
2) Needs to be massively tested and see that we did not cause any regressions with the new code
    (only minimal testing has been performed so far)

Thanks,
Ofir

ofirc:
References:
[1] INF directives
https://docs.microsoft.com/en-us/windows-hardware/drivers/install/adding-software-components-with-an-inf-file
[2] POC
https://github.com/ofirc/OCLEnumerator

ofirc:
@bashbaug @kepatil 
Hi Guys,
Important sidenote:
Microsoft suggestion for RS3 is to make "Component INFs" for Vulkan and OpenCL.
OpenGL and Vulkan will stay in "Base INF", but migrate keys to HKR (instead of HKLM), similar to the requirement from OpenCL.

Now that it has been tested more thoroughly and approved by Microsoft, we can start with the **review process**.

Let's not merge this until it is fully tested and known not to cause regressions for older OSes. I will update further.

P.S. a forked version of this code will be checked in to the Khronos Vulkan ICD Loader repository.

Thanks,
Ofir

ofirc:
The changes are ready to go.

ofirc:
@bashbaug @kepatil ?

icd_windows_hkr.c
```diff
@@ -0,0 +1,369 @@
...
+    do                              \
+    {                               \
+        free(mem);                  \
+        mem = NULL;                 \
+    } while (0)
+
+#ifdef _WIN64
+static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverName";
+#else
+static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverNameWow";
```
scyganINTEL: Typically, WoW refers only to executing 32-bit code on 64-bit OS. 

To preserve this semantics, it would be better to change this `#ifdef _WIN64`  to checking  for WoW with IsWow64Process() Windows API function.

ofirc: That's true in general, but in this case it is aligned with how OpenGL registers its ICDs, and uses the Wow suffix to refer to the 32-bit version of its ICD.
Besides, _WIN64 is compile-time, IsWow64Process() is run-time, we don't want the performance overhead incurred by this call if not actually needed.

kepatil:
@ofirc Changes to ICD extension spec were approved at the last working group call. I will act on this pull request shortly.

b-sumner:
I understand this is ready to go.  Kedar, would you please commit?  Our own tests show this is working as expected.

kepatil:
Relevant updates to the ICD extension were accepted in KhronosGroup/OpenCL-Registry#30.

PR #22, opened 2017-08-25
Add >VS2008 envvar

mirh:
[Idk](http://help.appveyor.com/discussions/kb/48-vs2017) about VS2017 then (which in other news also add native cmake support)

### DIFF
--- a/build_using_cmake.bat
+++ b/build_using_cmake.bat
@@ -1,4 +1,8 @@
 call "%VS90COMNTOOLS%/vsvars32.bat"
+call "%VS100COMNTOOLS%/vsvars32.bat"
+call "%VS110COMNTOOLS%/vsvars32.bat"
+call "%VS120COMNTOOLS%/vsvars32.bat"
+call "%VS140COMNTOOLS%/vsvars32.bat"
 
 set BUILD_DIR=build
 set BIN_DIR=bin


bashbaug:
I think this is fine, but would it be even better to build using cmake instead of invoking `nmake` directly?

Then, the batch file could simply become:

````
set BUILD_DIR=build
set BIN_DIR=bin

mkdir %BUILD_DIR%
cd %BUILD_DIR%
cmake ../
cmake --build .
cd ..
````

mirh:
Mh, well, that works too. 
I wouldn't be sure about pros and cons of the thing though. 
For example, a funny difference I noticed (aside from cmd spawning VS compiler logs, instead of a "gcc-like" progress screen) is that the exe manifest uses `'` instead of `"` now. 
This should be insignificant, but makes me wonder what's else.

bashbaug:
Are the differences purely cosmetic?  I can't think of any good reason why one method would be any more right than the other, just different.

To be clear, I'm OK with the proposed changes as-is, though it would be nice to avoid needing an update for each new VS version.

@kepatil any thoughts?

kepatil:
> Are the differences purely cosmetic? I can't think of any good reason why one method would be any more right than the other, just different.
> 
> To be clear, I'm OK with the proposed changes as-is, though it would be nice to avoid needing an update for each new VS version.
> 
> @kepatil any thoughts?

I think building using cmake as you suggested is better. I'll make that change.

bashbaug:
We've removed the batch files and switched entirely to a CMake build system in #61, so I am closing this pull request.

PR #26, opened 2017-11-23
Define build environment while calling make - Linux

PSGXerus:
This is an update to the existing Makefile, so users can specify their build environment while calling make. It is very handy, if this repository is used as a git submodule of a bigger project, because no patches need to be made on CMakeLists and one can use the vanilla master branch.
A description has also been added to the Linux part of the readme.

Specifying the BUILD_DIR can be dangerous whilst calling clobber. This risk is reduced by having the Makefile check the existence of CMakeCache.txt in the directory to be cleaned.


kepatil:
@PSGXerus Can you sign the CLA so that I can accept this change?

PSGXerus:
Done. Thanks for merging.

Makefile
```diff
@@ -26,4 +34,8 @@ clean:
 	${MAKE} -C ${BUILD_DIR} clean
 
 clobber:
+ifeq ($(wildcard ${BUILD_DIR}/CMakeCache.txt),)
+	$(error CMakeCache.txt does not exist in the specified BUILD_DIR. You might be pointing \
+	to the wrong directory to be cleaned or you have not even built the sources yet)
+endif
```
kepatil: This will potentially break workflows that unconditionally do "make clobber" before a build to ensure a clean build from scratch. This could perhaps be a warning or just a plain status message instead of an error.

PSGXerus: This is tricky, because I give the user the opportunity to specify BUILD_DIR. If a user would somehow specify BUILD_DIR as /, e. g. "make BUILD_DIR=/ clobber", he would wipe his system. And with the beforehand solution, "make -f some/dir/in/system/OpenCL-ICD-Loader" would also have this devastating result.

PSGXerus: I try to come up with a solution

PSGXerus: I changed the assertion to a warning and added a prompt to the user. Changes are ready for review.

kepatil:
@bashbaug I think we should work on #61 instead of merging this one since your change will benefit all cmake-supported platforms, including linux. Do you have any thoughts otherwise?

bashbaug:
> I think we should work on #61 instead of merging this one since your change will benefit all cmake-supported platforms, including linux. Do you have any thoughts otherwise?

This sounds reasonable to me.  @PSGXerus, will the changes in #61 work for you?

Looks like there are some conflicts in my PR now.  I'll rebase it and remove WIP by EOD.  Thanks!

PSGXerus:
@kepatil , @bashbaug : Sounds good. Mine was just a fix, but CMake is a proper solution.

PR #27, opened 2018-01-05
Fix dangling pointer in driver stub

ISilence:
The patch dealing with some nasty bugs in driver stub to pass failed tests.


bashbaug:
Looks good to me.  This fixes a warning in the ICD loader as well:

````
.../test/driver_stub/cl.c:182:17: warning: assignment from incompatible pointer type [-Wincompatible-pointer-types]
````

PR #28, opened 2018-02-22
Fix vendors enumeration on 32-bit Windows OS

ofirc:
When running 32-bit OpenCL applications on a 32-bit OS, we need to
use the registry keys without the "Wow" suffix. On 64-bit OSes and
32-bit applications, OTOH, we must use the Wow suffix.

### DIFF
--- a/icd_windows_hkr.c
+++ b/icd_windows_hkr.c
@@ -63,11 +63,30 @@ typedef enum
         mem = NULL;                 \
     } while (0)
 
-#ifdef _WIN64
 static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverName";
-#else
-static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverNameWow";
+
+#ifndef _WIN64
+static const char OPENCL_REG_SUB_KEY_WOW[] = "OpenCLDriverNameWow";
 #endif
+
+// Do not free the memory returned by this function.
+static const char* GetOpenCLRegKeyName(void)
+{
+#ifdef _WIN64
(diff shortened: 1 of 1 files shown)


icd_windows_hkr.c
```diff
@@ -63,11 +63,30 @@ typedef enum
         mem = NULL;                 \
     } while (0)
 
-#ifdef _WIN64
 static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverName";
+
+#ifndef _WIN64
```
kepatil: This should be ifdef instead of ifndef, right?

Same in the function below.

ofirc:
@kepatil Any updates :)?

kepatil:
Sorry, I did not notice this comment. @ofirc, can you address the comment I made on your change?

ofirc:
I already addressed them above:
1) ifndef
   The SUB_KEY, OpenCLDriverNameWow, is relevant only to 32-builds so this #ifndef takes care to avoid defining this variable from 64-bit builds (which results in a smaller footprint of OpenCL.dll).

2) ifdef
On 64-bit builds of OpenCL.dll, OTOH, the only valid name is the OPENCL_REG_SUB_KEY, i.e. OpenCLDriverName, so we return earliest as possible (as no checks are needed).

Everything that can be deferred to compile time is preferred, and every byte that is not needed is saved.

kepatil:
Okay. For some reason I your comments on the change were not visible to me till you posted them as a separate comment. The change looks good to me after your explanation. I had gotten the logic backwards in my mind. I'll merge this.

PR #30, opened 2018-04-05
Remove GL/EGL headers

ltowarek:
Remove GL/EGL headers from the ICD and replace GL types with typedefs from cl_platform.h.

The ICD requires only 3 GL types for CL/GL sharing related functions:
- GLenum
- GLint
- GLuint

All of these types are already mirrored in [cl_platform.h](https://github.com/KhronosGroup/OpenCL-Headers/blob/master/opencl22/CL/cl_platform.h#L409), so there is no point in including GL/EGL again. This way GL/EGL headers are no more required to build the ICD.

What's more, [clCreateFromGLTexture](https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/master/icd_dispatch.h#L807) is already using the mirrored types.


kepatil:
Thanks @ltowarek! This was on my list for a long time but I never got around to doing it.

PR #33, opened 2018-05-18
Windows 10 RS4 support for OpenCL ICD as per MS directive

saleelk:
These are set of changes required by MS for Para-Virtualization support for Windows 10 RS4


kepatil:
@saleelk Thanks for the patch! Can you add a reference to the relevant Microsoft documentation to the change description? It is not possible to review your change meaningfully without knowing what directives you are targeting.

kepatil:
Ping @saleelk for my request above. Also, please sign the CLA to allow me to accept your changes.

saleelk:
CLA signed

saleelk:
@kepatil The documentation for that change is under NDA. I am wondering each vendor must have received similar documentation from MS.

icd_windows_hkr.c
```diff
@@ -63,30 +63,11 @@ typedef enum
...
-    // application is running on a 64-bit Windows OS. A 32-bit application
-    // running on a 32-bit OS uses non-WoW names.
-    BOOL is_wow64;
-    if (IsWow64Process(GetCurrentProcess(), &is_wow64) && is_wow64)
-    {
-        return OPENCL_REG_SUB_KEY_WOW;
-    }
-
-    return OPENCL_REG_SUB_KEY;
+static const char OPENCL_REG_SUB_KEY[] = "OpenCLDriverNameWow";
```
kepatil: The logic that got removed from here is needed for 32 bit processes running on 64 bit windows.

saleelk: Looks like something got added to the master after I did a pull request. The file needs to be appended. Only REG_SZ needs to be changed to REG_MULTI_SZ here.

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+#if defined(DXGKDDI_INTERFACE_VERSION_WDDM2_4) && (DXGKDDI_INTERFACE_VERSION >= DXGKDDI_INTERFACE_VERSION_WDDM2_4)
+    {
+        D3DKMT_ADAPTERINFO* pAdapterInfo;
+        D3DKMT_ENUMADAPTERS2 EnumAdapters;
+        NTSTATUS Status = STATUS_SUCCESS;
+
+        // Get handle to GDI Runtime
+        HMODULE h = LoadLibrary("gdi32.dll");
+
+        char cszLibraryName[1024] = { 0 };
```
ofirc: Do we want MAX_PATH here?

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+        EnumAdapters.NumAdapters = 0;
+        EnumAdapters.pAdapters = NULL;
+        PFND3DKMT_ENUMADAPTERS2 pEnumAdapters2 = (PFND3DKMT_ENUMADAPTERS2)GetProcAddress((HMODULE)h, "D3DKMTEnumAdapters2");
+        if (!pEnumAdapters2)
+        {
+            FreeLibrary(h);
```
ofirc: We probably want a cleanup handler with C-style goto's so that we DRY (do not repeat ourselves).

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+            return FALSE;
+        }
+        Status = pEnumAdapters2(&EnumAdapters);
+        if (!NT_SUCCESS(Status) && (Status != STATUS_BUFFER_TOO_SMALL))
```
ofirc: Do we want an || or an && here?

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+            FreeLibrary(h);
+            return FALSE;
+        }
+        pAdapterInfo = (D3DKMT_ADAPTERINFO*)malloc(sizeof(D3DKMT_ADAPTERINFO)*(EnumAdapters.NumAdapters));
+        EnumAdapters.pAdapters = pAdapterInfo;
+        Status = pEnumAdapters2(&EnumAdapters);
+        if (!NT_SUCCESS(Status))
+        {
+            FreeLibrary(h);
+            if (pAdapterInfo) free(pAdapterInfo);
```
ofirc: free can be called with a null pointer, no need to check :-)

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+        for (UINT AdapterIndex = 0; AdapterIndex < EnumAdapters.NumAdapters; AdapterIndex++)
+        {
+            D3DDDI_QUERYREGISTRY_INFO QueryArgs = {0};
+            D3DDDI_QUERYREGISTRY_INFO* pQueryArgs = &QueryArgs;
+            D3DDDI_QUERYREGISTRY_INFO* pQueryBuffer = NULL;
+            QueryArgs.QueryType = D3DDDI_QUERYREGISTRY_ADAPTERKEY;
+            QueryArgs.QueryFlags.TranslatePath = TRUE;
+            QueryArgs.ValueType = REG_MULTI_SZ;
+#ifdef _WIN64
+            wcscpy_s(QueryArgs.ValueName, ARRAYSIZE(L"OpenCLDriverName"), L"OpenCLDriverName");
```
ofirc: We already have a define for OpenCLDriverName somewhere we could reuse, don't we?

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+                Status = D3DKMTQueryAdapterInfo(&QueryAdapterInfo);
+                pQueryArgs = pQueryBuffer;
+            }
+            if (NT_SUCCESS(Status) && pQueryArgs->Status == D3DDDI_QUERYREGISTRY_STATUS_SUCCESS)
+            {
+                wchar_t* pWchar = pQueryArgs->OutputString;
+                unsigned int i = 0;
+                memset(cszLibraryName, 0, sizeof(cszLibraryName));
+                {
+                    while ((*pWchar != L'\0') && (i<1023))
```
ofirc: 1023 should be constified or better yet, sizeof(buffer)

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+                    {
+                        cszLibraryName[i] = (char)*pWchar;
+                        i++;
+                        pWchar++;
+                    }
+                    if (i < 1023) khrIcdVendorAdd(cszLibraryName);
```
ofirc: Same goes here for 1023, and we should put the khrIcdVendorAdd on the following line.

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+            }
+            if (NT_SUCCESS(Status) && pQueryArgs->Status == D3DDDI_QUERYREGISTRY_STATUS_SUCCESS)
+            {
+                wchar_t* pWchar = pQueryArgs->OutputString;
+                unsigned int i = 0;
+                memset(cszLibraryName, 0, sizeof(cszLibraryName));
+                {
+                    while ((*pWchar != L'\0') && (i<1023))
+                    {
+                        cszLibraryName[i] = (char)*pWchar;
```
ofirc: Why do we ignore the 2nd byte? It's a potentially unicode character, don't we need to account for it or use a multi-byte to character converter?

icd_windows_dxgk.c
```diff
@@ -0,0 +1,145 @@
...
+                        i++;
+                        pWchar++;
+                    }
+                    if (i < 1023) khrIcdVendorAdd(cszLibraryName);
+                }
+            }
+            if (pQueryBuffer) free(pQueryBuffer);
```
ofirc: free can be used for null pointers as well, no need to check.

icd_windows_hkr.c
```diff
@@ -359,6 +341,8 @@ bool khrIcdOsVendorsEnumerateHKR(void)
                     &szGuid,
                     0);
 
+                KHR_ICD_ASSERT(devpropType == DEVPROP_TYPE_GUID);
```
ofirc: If the ret is not success the assert seems a bit weird, as devpropType might not get populated the way we think it should.

ofirc:
By the way, 2 other comments I forgot:
1) We should add the new file (icd_windows_dxgk.c) to CMakeLists.txt for Windows only builds
2) IMHO for tracking and clean history of OpenCL-ICD-Loader repository we should
    mention the motivation for the change in the commit message, see for example this:
   https://github.com/KhronosGroup/OpenCL-ICD-Loader/commit/188be6047481ea3c0c2ba5a7583ee13c8c3b2333

   for more info.

Thanks,
Ofir

icd_windows_hkr.c
ofirc: I think this should be in a common file to both dxgk and the other file.

icd_windows_hkr.c
```diff
@@ -115,7 +96,7 @@ static bool ReadOpenCLKey(DEVINST dnDevNode)
     {
         result = RegQueryValueExA(
             hkey,
-            GetOpenCLRegKeyName(),
+            OPENCL_REG_SUB_KEY,
```
ofirc: We should use the GetOpenCLRegKeyName() here and also in icd_windows_hkr.c line 139, I missed that line on my previous patch.

nikhiljnv:
Can we address open issues here? There are other changes dependent on this change that will need to be pulled in soon.

icd_windows_dxgk.c
saleelk: Do you mean to just use wcstombs

icd_windows_dxgk.c
saleelk: Fixed

icd_windows_hkr.c
saleelk: Fixed

icd_windows_dxgk.c
saleelk: its && here, Its to ensure num of adapters didnt increase between between calls

ofirc: But what if it failed for some other reason, i.e. the Status is not STATUS_BUFFER_TOO_SMALL?
Shouldn't we catch it and fail as well?

icd_windows_dxgk.c
ofirc: Probably

ofirc: Also, better to use sizeof(cszLibraryName) than 1024, it'll be future proof for changes.

icd_windows_hkr.c
nikhiljnv: @saleelk,
Are we fixing this as part of this commit?

icd_windows_dxgk.c
saleelk: Yea, i catch it for anything other than BUFFER_TOO_SMALL. I'll send the changes for review soon. Actually we've found some other issues as well and I was waiting until the new changes were tested but need to check with the engineer.

icd_windows_dxgk.c
saleelk: Yup. Fixed that already.

loader/windows/icd_windows_dxgk.c
```diff
@@ -0,0 +1,120 @@
...
+#define STATUS_SUCCESS  ((NTSTATUS)0x00000000L)
+#define STATUS_BUFFER_TOO_SMALL ((NTSTATUS)0xC0000023)
+#define NT_SUCCESS(Status) (((NTSTATUS)(Status)) >= 0)
+#endif
+
+#include "d3dkmthk.h"
+
+bool khrIcdOsVendorsEnumerateDXGK(void)
+{
+#if defined(DXGKDDI_INTERFACE_VERSION_WDDM2_4) && (DXGKDDI_INTERFACE_VERSION >= DXGKDDI_INTERFACE_VERSION_WDDM2_4)
```
ofirc: I should it would be easier to read and maintain the code if we had all FreeLibrary and free() calls at the same place (label), and have all the failures "goto" there, and the fall-through for the success path as well.

out:
    if (h != NULL)
        FreeLibrary(h);
    free(pAdapterInfo);
    return Status;

loader/windows/icd_windows_dxgk.c
```diff
@@ -0,0 +1,120 @@
...
+            Status = D3DKMTQueryAdapterInfo(&QueryAdapterInfo);
+            if (!NT_SUCCESS(Status))
+            {
+                free(pAdapterInfo);
+                goto out;
+            }
+            if (NT_SUCCESS(Status) && pQueryArgs->Status == D3DDDI_QUERYREGISTRY_STATUS_BUFFER_OVERFLOW)
+            {
+                unsigned int QueryBufferSize = sizeof(D3DDDI_QUERYREGISTRY_INFO) + QueryArgs.OutputValueSize;
+                pQueryBuffer = (D3DDDI_QUERYREGISTRY_INFO*)malloc(QueryBufferSize);
```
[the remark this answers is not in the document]
saleelk: we  release it after we get the OCL library name, Line 110 above

loader/windows/icd_windows_dxgk.c
```diff
@@ -0,0 +1,120 @@
...
+            QueryAdapterInfo.PrivateDriverDataSize = sizeof(QueryArgs);
+            Status = D3DKMTQueryAdapterInfo(&QueryAdapterInfo);
+            if (!NT_SUCCESS(Status))
+            {
+                free(pAdapterInfo);
+                goto out;
+            }
+            if (NT_SUCCESS(Status) && pQueryArgs->Status == D3DDDI_QUERYREGISTRY_STATUS_BUFFER_OVERFLOW)
+            {
+                unsigned int QueryBufferSize = sizeof(D3DDDI_QUERYREGISTRY_INFO) + QueryArgs.OutputValueSize;
```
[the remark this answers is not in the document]
saleelk: Yup. Fixed.

loader/windows/icd_windows_dxgk.c
```diff
@@ -0,0 +1,120 @@
...
+            }
+            if (NT_SUCCESS(Status) && pQueryArgs->Status == D3DDDI_QUERYREGISTRY_STATUS_SUCCESS)
+            {
+                wchar_t* pWchar = pQueryArgs->OutputString;
+                unsigned int i = 0;
+                memset(cszLibraryName, 0, sizeof(cszLibraryName));
+                {
+                    size_t ret = wcstombs(cszLibraryName, pWchar, sizeof(cszLibraryName));
+                    KHR_ICD_ASSERT(ret == (sizeof(cszLibraryName) - 1));
+                    if (i < (sizeof(cszLibraryName) - 1)) khrIcdVendorAdd(cszLibraryName);
```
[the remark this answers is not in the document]
saleelk: Fixed. Remnant of prev logic

bashbaug:
Hi @saleelk , we discussed this PR in our weekly teleconference, and it looks like it is getting very close.  Thank you for your patience addressing our review comments!

I'm afraid I have one more, though, related to PR #33 (which, as I understand, was folded into this PR).  Do we need to use `REG_MULTI_SZ` to query ICDs in the registry?  If so, we'll need to update the `cl_khr_icd` extension spec, since currently it only searches for `REG_SZ` registry values:

> To enumerate Vendor ICDs on Windows, the ICD Loader will first scan for REG_SZ string values in the "Display Adapter" and "Software Components" HKR registry keys.

A few other administrative comments:

* It looks like there are two copies of the icd_windows_dxgk.c and icd_windows_dxgk.h files - one in the "loader" directory and one in the "loader/windows" directory.  I think the one in the "loader" directory can be deleted.  I can clean this up myself when merging, if needed, but it'll save me a step if you can do it yourself.

* These changes didn't build for me.  The error was: 

    > icd_windows_dxgk.c(31): fatal error C1083: Cannot open include file: 'd3dkmthk.h': No such file or directory

    Do we need to update the README with new build dependencies?  Is a change to the CMakefile missing?

* The CLA Assistant is still indicating an issue with the CLA.  I suspect this is because the changes were committed as two different users.  Can you try to fix this?  A rebase may help, and it looks like there are merge conflicts anyhow.

Thanks!

loader/windows/icd_windows_dxgk.c
```diff
@@ -0,0 +1,141 @@
...
+            goto out;
+        }
+        for (UINT AdapterIndex = 0; AdapterIndex < EnumAdapters.NumAdapters; AdapterIndex++)
+        {
+            D3DDDI_QUERYREGISTRY_INFO QueryArgs = {0};
+            D3DDDI_QUERYREGISTRY_INFO* pQueryArgs = &QueryArgs;
+            D3DDDI_QUERYREGISTRY_INFO* pQueryBuffer = NULL;
+            QueryArgs.QueryType = D3DDDI_QUERYREGISTRY_ADAPTERKEY;
+            QueryArgs.QueryFlags.TranslatePath = TRUE;
+            QueryArgs.ValueType = REG_MULTI_SZ;
```
bashbaug: Here is the check for `REG_MULTI_SZ`.  The ICD spec currently says that only `REG_SZ` registry values are searched.  Is there some reason why this code is using `REG_MULTI_SZ`?

saleelk: Just fixed it. Sorry I've accumulated various changes from our trunk, and majority of the confusion here is because OpenGL string is MULTI_SZ so our installer was creating that incorrectly.

saleelk:
A few other administrative comments:

It looks like there are two copies of the icd_windows_dxgk.c and icd_windows_dxgk.h files - one in the "loader" directory and one in the "loader/windows" directory. I think the one in the "loader" directory can be deleted. I can clean this up myself when merging, if needed, but it'll save me a step if you can do it yourself.

[SKUDCHAD] - Fixed

These changes didn't build for me. The error was:

icd_windows_dxgk.c(31): fatal error C1083: Cannot open include file: 'd3dkmthk.h': No such file or directory

Do we need to update the README with new build dependencies? Is a change to the CMakefile missing?
[SKUDCHAD] - It seems we need to have MS SDK and WDK for that file, so yes it would need CMakefile changes. This builds on our local trunk as we have those in the path

The CLA Assistant is still indicating an issue with the CLA. I suspect this is because the changes were committed as two different users. Can you try to fix this? A rebase may help, and it looks like there are merge conflicts anyhow.
[SKUDCHAD] - Yes

bashbaug:
Thank you for the fast response!

I can confirm that the CLA issue and the REG_MULTI_SZ issue are fixed.

I'm still seeing two copies of icd_windows_dxgk.c and icd_windows_dxgk.h, but I can deal with these, no problem.

I'm still unable to build an ICD loader for Windows after these changes though, and I think this needs to be fixed.  One option would be to wrap `#include "d3dkmthk.h"` in the same set of defines that control the contents of `khrIcdOsVendorsEnumerateDXGK`:

```c
#if defined(DXGKDDI_INTERFACE_VERSION_WDDM2_4) && (DXGKDDI_INTERFACE_VERSION >= DXGKDDI_INTERFACE_VERSION_WDDM2_4)
```

I think the intent is for most (all?) Windows ICD loader builds to include this code, though, so perhaps this isn't the best solution?

One more thought: should we bump the file version after this change?  CC @kepatil @nikhiljnv

saleelk:
Fixed the duplicates in the branch. Also fixed the 32bit logic for Regkey. I installed the latest Windows WDK on my machine and it installs that file. So we need a check in cmake file to check if that exists, It also means we need to update the README so that folks install that. What do you guys think?

saleelk:
The build passes now but there is a dependency for Windows WDK for building ICD. This needs to be documented.

saleelk:
I've made a change to the cmakefile so that it searches and adds the necessary includes from the Windows WDK. ICD thus depends on the latest Windows SDK/WDK. @bashbaug ,What do you think?

bashbaug:
> I've made a change to the cmakefile so that it searches and adds the necessary includes from the Windows WDK. ICD thus depends on the latest Windows SDK/WDK. @bashbaug, What do you think?

Apologies for the slow reply.  It looks like the automated builds are passing now, which is great, but I am not able to build an ICD loader on a system without the WDK installed, which is a bit problematic.  I think we need to decide two things:

1. Should we have a build dependency on the Windows WDK?
1. What should happen if the Windows WDK is not installed?

I don't have a strong opinion about (1).  For (2) though, I'd prefer to see either an early and informative error from the CMakefile if there is a hard dependency on the Windows WDK and it's not found, or a way to build successfully without the Windows WDK if there is only a soft dependency.  Right now, the first indication of a problem is a compiler error, which is not particularly informative:

    icd_windows_dxgk.c(31): fatal error C1083: Cannot open include file: 'd3dkmthk.h': No such file or directory

Are there best practices from other APIs (Vulkan?) that we can follow?

kepatil:
Closing in favor of #78.

PR #43, opened 2018-12-02
Modify cmake build scripts

coast77777777777:
Include another script for VS2017 x86 build using cmake


bashbaug:
We've removed the batch files and switched entirely to a CMake build system in #61, so I am closing this pull request.  Thanks!

PR #45, opened 2019-01-04
Added OPENCL_VISIBLE_DEVICES support

ghost:
Supported an OPENCL_VISIBLE_DEVICES environment variable inspired from CUDA_VISIBLE_DEVICES.

OPENCL_VISIBLE_DEVICES=descriptor_list
descriptor_list -> descriptor | descriptor::descriptor_list
descriptor -> driver:platform | driver:platform:type | driver:platform:type:device_list
device_list -> device_id | device_id,device_list
driver: client driver (foo.icd or foo.so)
platform: platform ID
type: gpu, cpu, accelerator, custom, or any
device_id: device ID in the platform

e.g., OPENCL_VISIBLE_DEVICES=foo.icd:0:gpu:0,1::bar.icd:1:cpu:0 only shows
      - GPU 0, 1 in the first platform of foo.icd
      - CPU 0 in the second platform of bar.icd

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,12 +5,12 @@ project (OPENCL_ICD_LOADER)
 set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set (CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
-set (OPENCL_ICD_LOADER_SOURCES icd.c icd_dispatch.c)
+set (OPENCL_ICD_LOADER_SOURCES icd.c icd_dispatch.c icd_envvar.c)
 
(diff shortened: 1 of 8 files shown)


Kerilk:
The issues I have with this pull-request are:
- it raises the complexity of functions that were previously simple wrappers,
- it modifies platforms functionalities, whereas other environment variables modify what/how platforms are presented to the user. Is modifying platform behavior in the scope of a platform loader?
 - It is partly redundant with above mentioned environment variables as it can be used to hide a platform.

I understand the purpose it serves and how it can seem appealing in different contexts:
 - can simplify/override device selection mechanism of applications,
 - can be used for resource allocation/arbitration in a multi-user context by hiding resources from users.

I am unsure the second point is correctly addressed with these kind of mechanisms, as system based mechanisms for resource arbitration seem more robust. And you don't address sharing at the sub-device level.

Brice

PR #46, opened 2019-01-14
Avoid Segmentation Fault in icd_loader_test if platform is NULL

dmikushin:


icd_dispatch.c
```diff
@@ -1442,8 +1442,9 @@ clGetExtensionFunctionAddressForPlatform(cl_platform_id platform,
 
     // fall back to vendor extension detection
 
-    // FIXME Now that we have a platform id here, we need to validate that it isn't NULL, so shouldn't we have an errcode_ret
-    // KHR_ICD_VALIDATE_HANDLE_RETURN_HANDLE(platform, CL_INVALID_PLATFORM);   
+    // Now that we have a platform id here, we need to validate that it isn't NULL.
+    KHR_ICD_VALIDATE_HANDLE_RETURN_ERROR(platform, CL_INVALID_PLATFORM);
```
kepatil: This is incorrect since this macro will return error code instead of returning a pointer. In the master branch right now we have the following line instead which works correctly:

1451     KHR_ICD_VALIDATE_HANDLE_RETURN_ERROR(platform, NULL);

So I am inclined to drop this pull request.

PR #51, opened 2019-02-21
Update icd_windows.c

cos-public:
Return TRUE from khrIcdOsVendorsEnumerate() even if RegOpenKeyExA() fails to open the key.


cos-public:
If RegOpenKeyExA fails, khrIcdOsVendorsEnumerate() returns FALSE which in turn tells InitOnceExecuteOnce() that it wasn't initialized properly. So it will be called over and over again on each call to OpenCL API. RegOpenKeyExA() may fail even if devices were found with khrIcdOsVendorsEnumerateHKR() function (which is the preferred way AFAIK) and "SOFTWARE\Khronos\OpenCL\Vendors" key is not expected to exist with latest drivers.

bashbaug:
LGTM.

PR #64, opened 2019-05-08
Downstream MinGW fixes

### LINKED ISSUE #11: Error when building OpenCL ICD Loader using MinGW
Compilation of OpenCL ICD Loader using MinGW gcc fails. Compiler complains about missing `InitOnceExecuteOnce` function, then linker complains about missing dl lib. More details about this problem and workaround needed to get libOpenCL.a are in [this thread on StackOverflow](https://stackoverflow.com/questions/39179957/how-to-work-around-the-lack-of-initonceexecuteonce-in-mingw-or-mingw-w64-gcc/41944508). Please fix this.

BTW, please also add some option to easily get static lib instead of shared one. I am going to link this lib with app used for computations on one of BOINC projects. It is easier to link everything statically and distribute single binary than having to provide additional libs.

mmha:
* Don't include DX SDK at all on MinGW (MinGW have their own headers)
* Lower case header name for cross-compilation on Linux

This PR does **not** fix #11, but contains the fixes required on our side to make it build again once upstream problems are worked around or fixed.

See also [this upstream mailing list thread](https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/CAJ127VwoVEXkZETwzowxXw-fK2O6sLmZF-6Si-7rM5E4q-efCw%40mail.gmail.com/).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@ if (WIN32)
     # Only add the DXSDK include directory if the environment variable is
     # defined.  Since the DXSDK has merged into the Windows SDK, this is
     # only required in rare cases.
-    if (DEFINED ENV{DXSDK_DIR})
+    if (DEFINED ENV{DXSDK_DIR} AND NOT (MINGW OR MSYS OR CYGWIN))
         include_directories ($ENV{DXSDK_DIR}/Include)
     endif ()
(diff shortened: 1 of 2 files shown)


bashbaug:
LGTM - merging.

PR #61, opened 2019-04-17
CMake Build System

bashbaug:
This is a WIP pull request for discussion.  It is one way to resolve #55.  It removes the build batch files and Makefiles and instead describes how to build using CMake in the README.

PR #69, opened 2019-05-12
Add references to C++ packages

mmha:
This will conflict with #61, but I'd like to start the discussion and fix the merge conflict on this PR later on.

A big part of usability is to get the library running inside of the development environment fast and easy. Even though the Khronos Group does not maintain the packages for the two big C++ package managers, we should give developers a hint on how they can integrate OpenCL into their project using vcpkg or Conan + CMake.

Who wants to read build scripts and search the internet for the availability of a package to find out how to consume a library?

For reference:
* As of now, I am maintaining the Conan package
* As of now, @MathiasMagnus is maintaining the vcpkg package

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,21 @@
+# Getting the ICD Loader
+
+<table>
+    <tr><th>Conan</th><th>vcpkg</th><th>CMake (v3.0 and up)</th></tr>
+    <tr>
+        <td><a href="https://bintray.com/bincrafters/public-conan/khronos-opencl-icd-loader%3Abincrafters/_latestVersion"><img src="https://api.bintray.com/packages/bincrafters/public-conan/khronos-opencl-icd-loader%3Abincrafters/images/download.svg"></img></a></td>
+        <td><a href="https://github.com/microsoft/vcpkg/tree/master/ports/opencl"><img src="https://img.shields.io/badge/vcpkg-opencl-blueviolet.svg"></img></a></td>
+        <td>
+<pre language="cmake"><code>
+find_package(OpenCL REQUIRED NO_MODULE)
+target_link_libraries(app PRIVATE OpenCL::OpenCL)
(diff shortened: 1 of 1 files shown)


bashbaug:
This seems like a reasonable thing to do.  I'd personally prefer to put the links to C++ packages towards the bottom of the README vs. the top, but I could make a case either way,

Note that the README has changed considerably, so this PR needs to be updated.

MathiasMagnus:
@mmha Apologies for having overlooked this contribution, it somehow fell off my stack.

With the "recent" developments around helping users more easily put together a development environment, I think this content would make most sense in the OpenCL-Guide under the CMake Build System Support chapter, specifically the parts detailing [package manager support](https://github.com/KhronosGroup/OpenCL-Guide/blob/main/chapters/cmake_build-system_support.md#package-manager-support). The current readme [already](https://github.com/KhronosGroup/OpenCL-ICD-Loader/tree/5b280f69e8312c41310a72c28c66cf24830ee85a#build-instructions) "gently" suggests users to use the [OpenCL-SDK](https://github.com/KhronosGroup/OpenCL-SDK) to obtain a complete developer package including the C++ headers (and utility libs by default) too. So I think the overarching issue you're trying to remedy here has been triaged. Let us know if you think so too.

FWIW I am a Khronos member (as things stand currently) so in Vcpkg land you do already have a 1st party ally. 😀 One of the first things I wanted to do after we shipped the [new batch of native samples and utility libraries](https://www.khronos.org/blog/your-opencl-developer-experience-just-got-upgraded) for the SDK was hook it all up into Vcpkg, but unfortunately the `vcpkg_from_github` command can't clone recursively, so I've [opened an issue](https://github.com/microsoft/vcpkg/issues/24521) asking for this feature to simplify the Vcpkg portfile. The SDK likely also simplifies things on the Conan end.

Issue #67, opened 2019-05-11
Add .clang-format


mmha:
We should enable consistent and testable formatting with a .clang-format.

I'm not particularly picky on the chosen style, as long as we define one :)

PR #70, opened 2019-05-12
Add .clang-format

mmha:
This is one possible way of resolving #67. This PR is pretty much just `clang-format -style=llvm -dump-config > .clang-format`.

In other words, I'm adopting the LLVM style here. If someone has strong opinions on certain options I'm happy to change them, but my thinking here was that by adopting a predefined style we can keep the bikeshedding to a minimum.

Once adopted, I can create a second PR to actually perform the reformat.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,126 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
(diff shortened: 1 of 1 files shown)


bashbaug:
Seems reasonable to me.  I agree that it's more important to have consistent formatting than it is to have a specific formatting.

It doesn't need to be done as part of this commit, but as part of the PR to perform the reformat could you also:

* Update the README with brief instructions about formatting?
* Update the CI files to flag improperly formatted code?

Thanks!

Issue #48, opened 2019-01-27
Environment Variable Support


bashbaug:
There have been a few issues requesting support for various OpenCL ICD loader environment variables.  For example:

* https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/31
* https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/4

We've seen similar requests on the Intel OpenCL forums, such as:

* https://software.intel.com/en-us/forums/opencl/topic/802064

I know these issues have been closed in the past as "won't fix", but I believe we should reconsider this decision.  Note that the Vulkan loader provides support for various environment variables:

* https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#table-of-debug-environment-variables

I'd like to see OpenCL ICD loader environment variable support at least for loading ICDs from a location other than `/etc/OpenCL/vendors`, and for controlling debug output to help root-cause installation problems.  Are there best practices from the Vulkan loader that we could apply to add similar functionality to the OpenCL ICD loader?

bensallen:
I agree with @bashbaug in this request. Hard coding the location to /etc/OpenCL/vendors makes alternative path installs difficult, eg. installing to a shared filesystem.

Currently we are hex-editing the loader in Intel's OpenCL SDK to make the install suitable for our environment. This obviously is not particularly ideal.

mmha:
Here's another use case: I intend to submit [the headers](https://github.com/mmha/conan-opencl-headers), [the ICD](https://github.com/mmha/conan-opencl-icd-loader) and cl.hpp (WIP) to [Bincrafters](https://github.com/bincrafters/) for inclusion in their [Conan repository](https://bintray.com/bincrafters/public-conan/).

Having environment variable support is *essential* for the ICD loader to find OpenCL implementations that may be installed as a Conan package at a non-standard location.

This is most likely also required for all the Linux distributions that don't follow the FHS like Clear Linux, Nix and Guix.

bashbaug:
PR #77 was just merged, so this ICD loader now supports environment variables to enumerate ICDs from a location other than `/etc/OpenCL/vendors`.  Please see the [README](https://github.com/KhronosGroup/OpenCL-ICD-Loader#table-of-debug-environment-variables) for details.

I'd still like to add support for controlling debug output via environment variables but this has lower priority, I think.

[...]

PR #77, opened 2019-06-25
initial ICD loader environment variable support

bashbaug:
This change adds initial support for ICD loader environment variables, see #48.

The environment variable support was largely ported from the Vulkan ICD loader, so it uses the same `secure_getenv()` function when available.

I've added support for two environment variables for now, which I have documented in the README:

| Environment Variable              | Behavior            |  Example Format      |
|:---------------------------------:|---------------------|----------------------|

Additional environment variables could be added in the future, see for example #45.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,3 +116,12 @@ Please create a GitHub issue to report an issue or ask questions.
 
 Contributions to the OpenCL ICD Loader are welcomed and encouraged.
 You will be prompted with a one-time "click-through" CLA dialog as part of submitting your pull request or other contribution to GitHub.
+
+## Table of Debug Environment Variables
+
+The following debug environment variables are available for use with the OpenCL ICD loader:
+
+| Environment Variable              | Behavior            |  Example Format      |
+|:---------------------------------:|---------------------|----------------------|
+| OCL_ICD_FILENAMES                 | Specifies a list of additional ICDs to load.  The ICDs will be enumerated first, before any ICDs discovered via default mechanisms. | `export OCL_ICD_FILENAMES=libVendorA.so:libVendorB.so`<br/><br/>`set OCL_ICD_FILENAMES=vendor_a.dll;vendor_b.dll` |
+| OCL_ICD_VENDORS                   | On Linux and Android, specifies a directory to scan for ICDs to enumerate in place of the default `/etc/OpenCL/vendors'. |  `export OCL_ICD_VENDORS=/my/local/icd/search/path` |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,14 +41,20 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/loader/icd_cmake_config.h.in
 
 set (OPENCL_ICD_LOADER_SOURCES
     loader/icd.c
-    loader/icd_dispatch.c)
+    loader/icd.h
+    loader/icd_dispatch.c
+    loader/icd_dispatch.h
+    loader/icd_envvars.h
+    loader/icd_platform.h)
 
 if (WIN32)
     list (APPEND OPENCL_ICD_LOADER_SOURCES 
         loader/windows/icd_windows.c
+        loader/windows/icd_windows_dxgk.c
+        loader/windows/icd_windows_dxgk.h
         loader/windows/icd_windows_envvars.c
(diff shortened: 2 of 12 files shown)


bashbaug:
I've rebased these changes so there are no longer any merge conflicts.  Unless there are review comments, this PR is ready to go.

bashbaug:
Can somebody please review and merge these changes?  @nikhiljnv or @alycm perhaps?

README.md
```diff
@@ -115,4 +115,13 @@ Please create a GitHub issue to report an issue or ask questions.
...
\ No newline at end of file
+You will be prompted with a one-time "click-through" CLA dialog as part of submitting your pull request or other contribution to GitHub.
+
+## Table of Debug Environment Variables
+
+The following debug environment variables are available for use with the OpenCL ICD loader:
+
+| Environment Variable              | Behavior            |  Example Format      |
+|:---------------------------------:|---------------------|----------------------|
+| OCL_ICD_FILENAMES                 | Specifies a list of additional ICDs to load.  The ICDs will be enumerated first, before any ICDs discovered via default mechanisms. | `export OCL_ICD_FILENAMES=libVendorA.so:libVendorB.so`<br/><br/>`set OCL_ICD_FILENAMES=vendor_a.dll;vendor_b.dll` |
```
kepatil: In the ICD registry keys or vendor files we allow specifying paths. So I think we should support the same here and rename this variable to OCL_ICD_ENTRIES.

README.md
```diff
@@ -115,4 +115,13 @@ Please create a GitHub issue to report an issue or ask questions.
...
+| OCL_ICD_VENDORS                   | On Linux and Android, specifies a directory to scan for ICDs to enumerate in place of the default `/etc/OpenCL/vendors'. |  `export OCL_ICD_VENDORS=/my/local/icd/search/path` |
```
kepatil: If we allow paths in OCL_ICD_ENTRIES as suggested above, I think this is not needed anymore?

README.md
```diff
@@ -115,4 +115,13 @@ Please create a GitHub issue to report an issue or ask questions.
 ## Contributing
 
 Contributions to the OpenCL ICD Loader are welcomed and encouraged.
-You will be prompted with a one-time "click-through" CLA dialog as part of submitting your pull request or other contribution to GitHub.
\ No newline at end of file
+You will be prompted with a one-time "click-through" CLA dialog as part of submitting your pull request or other contribution to GitHub.
+
+## Table of Debug Environment Variables
```
kepatil: Is there a reason to call these as "debug" variables?

README.md
bashbaug: This environment variable supports full paths to ICDs as well.  I can update the table to show this if it would be helpful; it only uses file names without full paths right now for brevity.  For example:

| ... | Example Format |
|--|--|
|... | `export OCL_ICD_FILENAMES=/path/to/libVendorA.so:libvendorB.so`

The name of the environment variable was purposefully chosen to be compatible with the Vulkan loader, so I'd prefer not to change it unless there is a very good reason to do so.

README.md
bashbaug: No, I think we should support both options as they have different uses:

`OCL_ICD_VENDORS` still uses the .icd files, but it reads them from a different directory than `/etc/OpenCL/vendors`.  It is only valid on Linux and Android (and probably OSX as well, when using the ICD loader).

`OCL_ICD_FILENAMEs` is used to explicitly load the specified ICDs, without requiring any .icd files.  It works on all OSes.

README.md
bashbaug: The Vulkan loader calls them "debug environment variables":

https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#table-of-debug-environment-variables

Other ICD loaders simply call them "environment variables".

I have a slight preference for "debug environment variables", since they probably won't need to be set (and shouldn't be set!) for most end-user installations, but I don't have a very strong opinion, and I'd be fine changing to just "Table of Supported Environment Variables".

README.md
kepatil: I agree that it is good to be in sync with Vulkan loader.

README.md
kepatil: I agree with the rationale.

PR #78, opened 2019-06-27
Windows 10 RS4 Support for the OpenCL ICD Loader w/ Build Option

bashbaug:
This is a slightly enhanced version of PR #33.  Changes are:

* Rebased to the latest master branch.
* Added a CMake option to build without a WDK dependency, for users that are simply building OpenCL applications and need a lib to link with.  By default, the OpenCL ICD loader will require the WDK for Windows builds, because without it some OpenCL implementations may fail to enumerate.
* Documented the WDK dependency and CMake option in the README.

Mentioning @nikhiljnv and @saleelk for awareness.


nikhiljnv:
@bashbaug , Can we also bump up the version with this change?

saleelk:
Looks good to me.

bashbaug:
> @bashbaug , Can we also bump up the version with this change?

Nice catch - I'd missed that this wasn't a part of the initial PR.

I've also added a bit of infrastructure to a) make it easier to update this in the future, and b) make it easier to sync the file version based on CMake variables, which IMHO is the best long term solution.

bashbaug:
Assigning to Kedar for final review and merging, unless someone else gets to it first.  Thanks!

Issue #76, opened 2019-06-20
Improved Unit Testing Feedback


bashbaug:
I've been experimenting with alternate ways to test the OpenCL ICD loader, with the following goals in mind:

1. Use a modern unit testing framework.
1. Test as much as possible without needing to manually setup a stub ICD.
1. Automatically generate as much test content as possible.

I have something working and I'm looking for feedback before proceeding further, mostly because there is still some manual work needed to setup each test, and I won't spend the time doing this if it looks like things aren't moving in the right direction.

Some details:

* My work is staged here:

    https://github.com/bashbaug/OpenCL-ICD-Loader/tree/improved-testing

    Most of the new code is currently in the "test_new" directory.  I can create a WIP pull request if that is easier for review.

* I'm using Googletest for now, which I am most familiar with, but I can switch to a different testing framework if desired.

* I've managed to generate almost all of the "recorder" ICD, which I use for testing.  The "recorder" ICD simply records each of the arguments it received in memory before returning.  The test code can then ensure that the arguments that were recorded match the arguments that it passed.  This is very similar to the existing "stub" ICD, except it records the arguments in memory rather than in a file.

* I've managed to generate a "template" to simplify test development for each API, but some parts of the template still need to be filled in manually.  If anyone has ideas to completely generate each test I'd love to hear them!

* Things this won't test: any of the ICD discovery code, such as the code to scan the registry or /etc/OpenCL/vendors.

What I'm looking for:

* Is this helpful?  Should I keep going?

* Are there any suggested tweaks to the test "template" before I start adding the remaining APIs?

Aside: I think we should be able to automatically generate a bunch of the ICD loader code itself, similar to the way I've generated the recorder ICD and test template.  I plan to look at this at some point in the near future, but if someone else wants to take a look at it first, the gen_tester.py script could be a helpful place to start from.

Thanks!

Kerilk:
I like the way you are testing the loader here. I think I have an idea to completely automate the generation of 95% of the test.
I have a couple of questions though, and one remark. Questions:
 - Is there any reason to avoid generating a mockup ICD? This would allow testing a 100% of the API while staying generated. I imagine the problem would lie with Windows, as on unixes I don't see any obstacles.
 - We could also use the same approach to create a test layer to check cases where the call can't dispatch, and the layer mechanism.
 -  Do you mind if I pick-up where you left it?
Remark:
[...]

Issue #73, opened 2019-05-16
Reduce Calls to khrIcdInitialize


bashbaug:
I'm trying to figure out why the ICD loader is calling **khrIcdInitialize** in the places where it is, and if some of the calls can be eliminated.

For reference, the current list of OpenCL APIs that are calling **khrIcdInitialize** is:

* clGetPlatformIDs
* clGetPlatformInfo
* clGetDeviceIDs
* clCreateContext
* clCreateContextFromType
* clUnloadPlatformCompiler
* clGetExtensionFunctionAddressForPlatform
* clGetExtensionFunctionAddress
* clGetGLContextInfoKHR

I believe the following APIs should be removed from the list, for the specified reasons:

* clGetPlatformInfo: Requires a cl_platform_id, which must be obtained using clGetPlatformIDs.
* clGetDeviceIDs: Also requires a cl_platform_id.
* clCreateContext: Requires a cl_platform_id or a cl_device_id, which must be obtained using clGetPlatformIDs or clGetDeviceIDs.
* clUnloadPlatformCompiler: Requires a cl_platform_id.
* clGetExtensionFunctionAddressForPlatform: Requires a cl_platform_id.
* clGetGLContextInfoKHR: This function is not exported by the ICD loader, so it must be retrieved using clGetExtensionFunctionAddress or clGetExtensionFunctionAddressForPlatform.

Can the list of APIs that call **khrIcdInitialize** be reduced to:

* clGetPlatformIDs
* clCreateContextFromType
* clGetExtensionFunctionAddress

bashbaug:
If there are no objections, I will create a pull request to remove the calls to `khrIcdInitialize` from all APIs except:

* clGetPlatformIDs
* clCreateContextFromType
* clGetExtensionFunctionAddress

Thanks!

bashbaug:
This has been done for some time now.  Since I don't think there is anything more to do, I am closing this issue.

PR #79, opened 2019-07-02
Use QueryAdapter APIs as per MS directives

nikhiljnv:
Use QueryAdapter APIs as per MS directives

Update ICD loader to use QueryAdapter APIs
as per MS directives for paravirtualization.

This includes changes to -
1. Order OpenCL platforms consistent with Windows adapters,
   to honor user-settings for multi-adapter machines.
2. Fix OS version detection.
3. Some other minor fixes including -
     - Fix build failure in getting the address of D3DKMTQueryAdapterInfo.
     - Fix early return platform enumeartion if there are no entries found for old reg key
       under "HKLM\\SOFTWARE\\Khronos\\OpenCL\\Vendors" to allow add for the entries found
        using DXGK interface and HKR entries.
4. Update to README.md

Merge branch 'skudchad-ocl-wddm24'
of https://github.com/BKlamik-MSFT/OpenCL-ICD-Loader


loader/icd_dispatch.c
```diff
@@ -99,6 +99,17 @@ clGetDeviceIDs(cl_platform_id   platform,
 {
     // initialize the platforms (in case they have not been already)
     khrIcdInitialize();
+
+    // determine the platform to use from the device_type specified
```
bashbaug: This is technically OK since the spec says that behavior is implementation-defined when `platform` is `NULL`, but I do wonder if this is right direction.  In general, I'd like to "thin out" these aspects of the ICD loader, to the point where it could be automatically generated in many cases (see: #73, #76).  Also, not that if we added this now, there would be some ICD loaders that return NULL and an error with a NULL platform, and others where things magically work, which seems undesirable.

Unless there is a strong reason to add this behavior, I'd prefer to revert this change.

loader/windows/icd_windows.c
```diff
@@ -17,14 +17,67 @@
  */
 
 #include "icd.h"
+#include <windows.h>
```
bashbaug: Requiring a specific order for #include files is almost always an indication that something is wrong.  If icd_windows_hkr.h or icd_windows_dxgk.h requires windows.h then they should include it themselves, vs. requiring windows.h to be included before them.

loader/windows/icd_windows.c
```diff
@@ -17,14 +17,67 @@
...
+    {
+        size_t OldCapacity = pWinAdapterCapacity - pWinAdapterBegin;
+        size_t NewCapacity = OldCapacity;
+        if (0 == NewCapacity)
+        {
+            NewCapacity = 1;
+        }
+        NewCapacity *= 2;
+
+        WinAdapter* pNewBegin = malloc(NewCapacity * sizeof(*pWinAdapterBegin));
```
bashbaug: Where is this memory freed?

loader/windows/icd_windows.c
nikhiljnv: Added windows.h in icd_windows_hkr.h.

loader/windows/icd_windows.c
nikhiljnv: Added free after enumerating WinAdapter entries.

loader/icd_dispatch.c
nikhiljnv: Reverted the part of the change which was overriding platform with platform[0].

nikhiljnv:
@BKlamik-MSFT,
Please, sign CLA for this pull request.

loader/icd_dispatch.c
bashbaug: I still see a change here that I'd prefer to revert.

loader/windows/icd_windows_hkr.c
```diff
@@ -17,8 +17,9 @@
  */
 
 #include "icd.h"
-#include "icd_windows_hkr.h"
 #include <windows.h>
+#include "icd_windows_hkr.h"
```
bashbaug: This change can be reverted now, since icd_windows_khr.h properly includes Windows.h,

loader/windows/icd_windows.c
bashbaug: Great, we change the include file order back here then.

nikhiljnv:
Closing this in favor of #81.

PR #80, opened 2019-07-02
Use API function pointer and dispatch table definitions from the CL headers

### LINKED ISSUE #72: Move the dispatch table declaration to OpenCL-Headers?
The various function pointer types as well as the declaration for the dispatch table in `icd_dispatch.h` would be easier to reuse directly in OpenCL implementations if they were part of OpenCL-Headers.

It seems to me that only the declaration for the structure types at the bottom of the file should be part of the ICD loader itself.

Is there a good reason to not do that?

kpet:
Depends on https://github.com/KhronosGroup/OpenCL-Headers/pull/50.

Fixes #72.

Signed-off-by: Kevin Petit <kevin.petit@arm.com>


kpet:
The headers change has gone in. I've rebased this and it's now passing the CI. I think this is now ready to go.

PR #81, opened 2019-07-12
Use QueryAdapter APIs as per MS directives

nikhiljnv:
Use QueryAdapter APIs as per MS directives

Update ICD loader to use QueryAdapter APIs
as per MS directives for para-virtualization.

This includes changes to -
1. Order OpenCL platforms consistent with Windows adapters,
   to honor user-settings for multi-adapter machines.
2. Fix OS version detection.
3. Some minor fixes including
    1. Add trace prints at required places.
    2. Fix build failure in getting the address of D3DKMTQueryAdapterInfo.
    3. Fix early return platform enumeartion if there are no entries found for old reg key
       under "HKLM\\SOFTWARE\\Khronos\\OpenCL\\Vendors" to allow add for the entries found
       using DXGK interface and HKR entries.
4. Some other fixes including
    1. Update README.md with WDK dependency information.
    2. Move AddAdapter* call inside for which got misplaced while resolving conflicts.
    3. Free WinAdapter allocations and avoid overriding NULL platform.
5. Fixes based on review comments including -
    1. WinAdapter is reallocated with previous allocation freed.
       However, the final allocation is not freed.
       Fixed the same to free at the end of enumeration.
    2. The changes for para-virtualization added overriding the
       platform with platform[0]. Revert the changes to keep
       existing behavior as is.
    3. Fix header file ordering by including windows headers
       inside necessary header files.


nikhiljnv:
The changes reverted as part of 5.ii need to be reconsidered as they contradict with 1.
The idea is to follow the Windows adapter ordering for enumeration. This also means that platform 0 needs to be initialized with the platform belonging to first adapter in the list.
@bashbaug , Any concerns? Does this require an update to ICD loader spec or guidelines?

loader/icd.h
```diff
@@ -121,6 +121,11 @@ void khrIcdContextPropertiesGetPlatform(
     const cl_context_properties *properties, 
     cl_platform_id *outPlatform);
 
+// translate device type to platform
+void khrIcdDeviceTypeGetPlatform(
```
bashbaug: Same comment as with https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/79: let's include this as part of a separate PR.  It doesn't look like it does anything right now, anyhow.

loader/windows/icd_windows_hkr.c
```diff
@@ -17,8 +17,9 @@
  */
 
 #include "icd.h"
-#include "icd_windows_hkr.h"
 #include <windows.h>
+#include "icd_windows_hkr.h"
```
bashbaug: We fixed the include file dependencies in #79, correct?  If so, let's keep the #include for <windows.h> where it was.

loader/windows/icd_windows_hkr.h
```diff
@@ -17,5 +17,13 @@
  */
 
 #include <stdbool.h>
+#include <windows.h>
 
 bool khrIcdOsVendorsEnumerateHKR(void);
+
+LUID ZeroLuid;
```
bashbaug: Should this be "extern"?

loader/icd.h
nikhiljnv: Removed.

loader/windows/icd_windows_hkr.c
nikhiljnv: Fixed.

loader/icd_dispatch.c
```diff
@@ -99,6 +99,12 @@ clGetDeviceIDs(cl_platform_id   platform,
 {
     // initialize the platforms (in case they have not been already)
     khrIcdInitialize();
+
+    if (!platform && khrIcdVendors != NULL)
```
bashbaug: These changes for a `NULL` platform are fine according to the spec, but this will mean that different versions of the ICD loader will have different behavior for a `NULL` platform.  I'd be more comfortable discussing them as part of a separate PR.

loader/icd_dispatch.c
```diff
@@ -196,8 +202,12 @@ clCreateContextFromType(const cl_context_properties * properties,
     // initialize the platforms (in case they have not been already)
     khrIcdInitialize();
 
-    // determine the platform to use from the properties specified
+    // determine the platform to use from the properties and device_type specified
```
bashbaug: Minor: This comment may have been from an earlier commit and the platform is no longer determined by the device type (if it ever was).

loader/windows/icd_windows_hkr.h
bashbaug: I still think this should be "extern".

nikhiljnv: Fixed.

loader/icd_dispatch.c
nikhiljnv: Removing for now to unblock rest of the changes. Let's discuss this separately.

loader/icd_dispatch.c
nikhiljnv: Fixed.

bashbaug:
All of my comments have been addressed - thanks!

I still think it would be wise to have an installer expert review these changes - @ofirc  or @saleelk perhaps?

ofirc:
@bashbaug I'll review it tomorrow

loader/windows/icd_windows.c
```diff
@@ -23,8 +23,61 @@
...
+                free(pWinAdapterBegin);
+            }
+            pWinAdapterCapacity = pNewBegin + NewCapacity;
+            pWinAdapterEnd = pNewBegin + OldCapacity;
+            pWinAdapterBegin = pNewBegin;
+        }
+    }
+    if (pWinAdapterEnd != pWinAdapterCapacity)
+    {
+        size_t nameLen = strlen(szName) + 1;
```
ofirc: use sizeof(szName[0]), if we ever change to unicode we won't have to change this line as well.

loader/windows/icd_windows.c
```diff
@@ -23,8 +23,61 @@
...
+        if (pWinAdapterEnd->szName = malloc(nameLen))
```
ofirc: And if not? do we still want this adapter?

loader/windows/icd_windows.c
```diff
@@ -103,18 +156,57 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
...
 
-        result = RegCloseKey(platformsKey);
-        if (ERROR_SUCCESS != result)
+    // Add adapters according to DXGI's preference order
+    HMODULE hDXGI = LoadLibrary("dxgi.dll");
+    if (hDXGI)
+    {
+        IDXGIFactory* pFactory = NULL;
+        PFN_CREATE_DXGI_FACTORY pCreateDXGIFactory = (PFN_CREATE_DXGI_FACTORY)GetProcAddress(hDXGI, "CreateDXGIFactory");
+        HRESULT hr = pCreateDXGIFactory(&IID_IDXGIFactory, &pFactory);
```
ofirc: Is this guaranteed to be present on any version of Windows?
If so add a: KHR_ASSERT(pCreateDXGIFactory != NULL);

loader/windows/icd_windows.c
```diff
@@ -103,18 +156,57 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
...
+        if (SUCCEEDED(hr))
         {
-            KHR_ICD_TRACE("Failed to close platforms key %s, ignoring\n", platformsName);
+            UINT i = 0;
+            IDXGIAdapter* pAdapter = NULL;
+            while (SUCCEEDED(pFactory->lpVtbl->EnumAdapters(pFactory, i++, &pAdapter)))
```
ofirc: Do we have to go through the table here and in line 178?
According to this you can call the method directly:
https://docs.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgiadapter-getdesc
IDXGIAdapter * pDXGIAdapter;
pDXGIDevice->GetAdapter(&pDXGIAdapter);

without needing to pass "this".

loader/windows/icd_windows.c
```diff
@@ -103,18 +156,57 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
...
+        FreeLibrary(hDXGI);
     }
 
+    // Go through the list again, putting any remaining adapters at the end of the list in an undefined order
+    for (WinAdapter* iterAdapter = pWinAdapterBegin; iterAdapter != pWinAdapterEnd; ++iterAdapter)
+    {
+        khrIcdVendorAdd(iterAdapter->szName);
+    }
+
+    free(pWinAdapterBegin);
```
ofirc: What about the rest of the list?

loader/windows/icd_windows_dxgk.c
```diff
@@ -83,6 +84,8 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
             KHR_ICD_TRACE("D3DKMT_ENUMADAPTERS2 status != SUCCESS\n");
             goto out;
         }
+        const char* cszOpenCLRegKeyName = GetOpenCLRegKeyName();
+        const int OpenCLRegKeyNameSize = (int)(strlen(cszOpenCLRegKeyName) + 1);
```
ofirc: 1) sizeof(cszOpenCLRegKeyName[0]);
2) Naming: use camelCase

loader/windows/icd_windows_dxgk.c
```diff
@@ -37,16 +37,17 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
     bool ret = false;
 #if defined(OPENCL_ICD_LOADER_REQUIRE_WDK)
 #if defined(DXGKDDI_INTERFACE_VERSION_WDDM2_4) && (DXGKDDI_INTERFACE_VERSION >= DXGKDDI_INTERFACE_VERSION_WDDM2_4)
+    // Get handle to GDI Runtime
+    HMODULE h = LoadLibrary("gdi32.dll");
+    if (h && GetProcAddress((HMODULE)h, "D3DKMTSubmitPresentBltToHwQueue")) // OS Version check
```
ofirc: 1) Resource leak
    We are not releasing `h` here, we should call FreeLibrary on the handle.
2) Redundant cast to HMODULE in GetProcAddress arg 0
3) Refactoring - there is a better way to express the intent here (IMHO)
I would move these two lines (predicate) into an inline function IsOSVersionGTE_XYZ:
static inline bool IsOSVersionGTE_XYZ(void)
{
    bool ret;
    HMODULE h = LoadLIbrary("gdi32.dll");
    ret = GetProcAddress(h, "D3DKMTSubmitPresentBltToHwQueue") != NULL;
    FreeLibrary(h);
    return;
}

loader/windows/icd_windows_dxgk.c
```diff
@@ -91,20 +94,13 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
-            if (IsWow64Process(GetCurrentProcess(), &is_wow64) && is_wow64)
-            {
-                wcscpy_s(QueryArgs.ValueName, ARRAYSIZE(L"OpenCLDriverNameWow"), L"OpenCLDriverNameWow");
-            }
-            else
-            {
-                wcscpy_s(QueryArgs.ValueName, ARRAYSIZE(L"OpenCLDriverName"), L"OpenCLDriverName");
-            }
-#endif
+            MultiByteToWideChar(
```
ofirc: Missing return value check.

loader/windows/icd_windows_dxgk.c
```diff
@@ -133,7 +130,7 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
                 {
                     size_t len = wcstombs(cszLibraryName, pWchar, sizeof(cszLibraryName));
```
ofirc: Where is pWchar defined?

loader/windows/icd_windows_hkr.h
```diff
@@ -17,5 +17,13 @@
  */
 
 #include <stdbool.h>
+#include <windows.h>
 
 bool khrIcdOsVendorsEnumerateHKR(void);
+
+extern LUID ZeroLuid;
```
ofirc: Constify it.

loader/windows/icd_windows.c
```diff
@@ -103,18 +156,57 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
...
+    }
 
-        result = RegCloseKey(platformsKey);
-        if (ERROR_SUCCESS != result)
+    // Add adapters according to DXGI's preference order
+    HMODULE hDXGI = LoadLibrary("dxgi.dll");
+    if (hDXGI)
+    {
+        IDXGIFactory* pFactory = NULL;
+        PFN_CREATE_DXGI_FACTORY pCreateDXGIFactory = (PFN_CREATE_DXGI_FACTORY)GetProcAddress(hDXGI, "CreateDXGIFactory");
```
saleelk: could just use this way instead? https://docs.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-createdxgifactory

loader/windows/icd_windows_dxgk.c
nikhiljnv: I think the gdi32 module is needed even after this check, so freeing it right after OS version check will need to reload it again. I would rather split gdi32 load and D3DKMTSubmitPresentBltToHwQueue query and move the query to a macro.

loader/windows/icd_windows_dxgk.c
nikhiljnv: Just above this block. See line 128.

nikhiljnv: > 
> 
> Where is pWchar defined?



Just above this block. See line 128.

ofirc:
Did we run this patch with a memory (leaks) checker?
I don't see where the adapters dynamic list (that we allocate in icd_windows.c) gets released.
As soon as the ICD Loader gets unloaded (FreeLibrary), we should release all associated resources with it (or at least try).
Ideally, we could register a callback with atexit() to release our global array.

loader/windows/icd_windows.c
```diff
@@ -103,18 +156,57 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
...
             // add the library
-            khrIcdVendorAdd(cszLibraryName);
+            AdapterAdd(cszLibraryName, ZeroLuid);
         }
+    }
 
-        result = RegCloseKey(platformsKey);
-        if (ERROR_SUCCESS != result)
+    // Add adapters according to DXGI's preference order
+    HMODULE hDXGI = LoadLibrary("dxgi.dll");
```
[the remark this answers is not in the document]
nikhiljnv: Yes, dxgi is required for CreateDXGIFactory and IDXGIAdapter.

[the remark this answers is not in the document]
saleelk: I was asking from the perspective of using D3DKMT_ENUMADAPTERS from gdi32

loader/windows/icd_windows_hkr.h
nikhiljnv: > 
> 
> Constify it.

Done!!

loader/windows/icd_windows.c
```diff
@@ -23,8 +23,61 @@
...
+    LUID luid;
+} WinAdapter;
+
+LUID ZeroLuid = { 0, 0 };
+
+static WinAdapter* pWinAdapterBegin = NULL;
+static WinAdapter* pWinAdapterEnd = NULL;
+static WinAdapter* pWinAdapterCapacity = NULL;
+
+void AdapterAdd(const char* szName, LUID luid)
```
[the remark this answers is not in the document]
nikhiljnv: > 
> 
> Naming inconsistency: we use camelCase for variables, change to: oldCapacity and newCapacity, etc

Fixed.

loader/windows/icd_windows.c
nikhiljnv: Fixed.

loader/windows/icd_windows_dxgk.c
nikhiljnv: Fixed.

loader/windows/icd_windows.c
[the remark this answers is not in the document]
nikhiljnv: DXGI honors the user order of adapters. See https://github.com/MicrosoftDocs/windows-driver-docs/pull/1538/files for more details.

loader/windows/icd_windows.c
```diff
@@ -23,8 +23,61 @@
...
+    {
+        size_t OldCapacity = pWinAdapterCapacity - pWinAdapterBegin;
+        size_t NewCapacity = OldCapacity;
+        if (0 == NewCapacity)
+        {
+            NewCapacity = 1;
+        }
+        NewCapacity *= 2;
+
+        WinAdapter* pNewBegin = malloc(NewCapacity * sizeof(*pWinAdapterBegin));
```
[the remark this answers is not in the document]
nikhiljnv: I would prefer doing this as part of separate merge-request.

loader/windows/icd_windows_dxgk.c
nikhiljnv: > 
> 
> Missing return value check.

Resolved.

loader/windows/icd_windows.c
nikhiljnv: The answer here depends on the choice of the default platform. 
I plan to discuss this as part of a separate follow up change.
See https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/81/#issuecomment-514270979 for earlier discussion on this.
Based on whether the default platform is already added to the list of adapters or not before we fail to allocate memory for adapters, we will need to either skip the adapter on out of memory or may need to evict some of the earlier adapters out of the list to make memory for the new adapter.

loader/windows/icd_windows_dxgk.c
nikhiljnv: I see this currently only at one place. So, I would defer this for now as there isn't a real gain from refactoring this. I am expecting few changes around this to follow.

loader/windows/icd_windows.c
```diff
@@ -23,8 +23,61 @@
...
+        if (pNewBegin)
```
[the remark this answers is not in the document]
nikhiljnv: I have changed the function to return boolean status to reflect success or failure.
However, to be able to propagate this back to API, we will need more changes as khrIcdVendorAdd is not returning the status. I would prefer to address this separately.

nikhiljnv:
@saleelk @ofirc ,
Can you take a look at latest patch? Let me know if there are any further comments/suggestions.

ofirc:
@nikhiljnv sorry for the delay, I'll review it tomorrow.

loader/windows/icd_windows_hkr.h
```diff
@@ -17,5 +17,13 @@
  */
 
 #include <stdbool.h>
+#include <windows.h>
 
 bool khrIcdOsVendorsEnumerateHKR(void);
+
+extern const LUID ZeroLuid;
+
+BOOL adapterAdd(const char* szName, LUID luid);
```
ofirc: We have inconsistency with the naming, either camelCase or PascalCase: getOpenCLRegKeyName() / AdapterAdd / AddAdapter(const char*...)

loader/windows/icd_windows.c
```diff
@@ -23,8 +23,75 @@
...
+            pWinAdapterEnd->luid = luid;
+            ++pWinAdapterEnd;
+        }
+    }
+    return result;
+}
+
+void adapterFree(WinAdapter *pWinAdapter)
+{
+    if(pWinAdapter->szName)
```
ofirc: You don't need to check for null, free is guaranteed to work according to the C standard.

loader/windows/icd_windows.c
```diff
@@ -103,19 +171,62 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
                 KHR_ICD_TRACE("Value not zero, skipping\n");
                 continue;
             }
-
             // add the library
-            khrIcdVendorAdd(cszLibraryName);
+            status |= adapterAdd(cszLibraryName, ZeroLuid);
```
ofirc: I'm not sure about the semantics here.
Do we want to set a valid status if *any* of the adapter were successfully added or iff all of them succeeded? (i.e. OR vs AND).

loader/windows/icd_windows.c
```diff
@@ -103,19 +171,62 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
...
+	{
+            HRESULT hr = pCreateDXGIFactory(&IID_IDXGIFactory, &pFactory);
+            if (SUCCEEDED(hr))
+            {
+                UINT i = 0;
+                IDXGIAdapter* pAdapter = NULL;
+                while (SUCCEEDED(pFactory->lpVtbl->EnumAdapters(pFactory, i++, &pAdapter)))
+                {
+                    DXGI_ADAPTER_DESC AdapterDesc;
+                    pAdapter->lpVtbl->GetDesc(pAdapter, &AdapterDesc);
```
ofirc: The return value is ignored here.

loader/windows/icd_windows_dxgk.c
```diff
@@ -83,47 +88,47 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
-            D3DKMT_QUERYADAPTERINFO QueryAdapterInfo = {0};
-            QueryAdapterInfo.hAdapter = pAdapterInfo[AdapterIndex].hAdapter;
-            QueryAdapterInfo.Type = KMTQAITYPE_QUERYREGISTRY;
-            QueryAdapterInfo.pPrivateDriverData = &QueryArgs;
-            QueryAdapterInfo.PrivateDriverDataSize = sizeof(QueryArgs);
-            Status = D3DKMTQueryAdapterInfo(&QueryAdapterInfo);
+            queryArgs.QueryType = D3DDDI_QUERYREGISTRY_ADAPTERKEY;
+            queryArgs.QueryFlags.TranslatePath = TRUE;
+            queryArgs.ValueType = REG_SZ;
+	    result = MultiByteToWideChar(
```
ofirc: Indentation inconsistency.

loader/windows/icd_windows_dxgk.c
```diff
@@ -83,47 +88,47 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
+                	CP_ACP,
+                	0,
+                	cszOpenCLRegKeyName,
+                	szOpenCLRegKeyName,
+                	queryArgs.ValueName,
+                	ARRAYSIZE(queryArgs.ValueName));
+	    if (!result) {
```
ofirc: ditto

loader/windows/icd_windows_dxgk.c
```diff
@@ -83,47 +88,47 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
             KHR_ICD_TRACE("D3DKMT_ENUMADAPTERS2 status != SUCCESS\n");
             goto out;
         }
+        const char* cszOpenCLRegKeyName = GetOpenCLRegKeyName();
+        const int szOpenCLRegKeyName = (int)(strlen(cszOpenCLRegKeyName) + 1)*sizeof(cszOpenCLRegKeyName[0]);
```
ofirc: the 1 can be replaced with sizeof(cszOpenCLRegKeyName[0])) for (possible) future compatibility with unicode.

loader/windows/icd_windows_dxgk.c
```diff
@@ -83,47 +88,47 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
+            queryArgs.QueryType = D3DDDI_QUERYREGISTRY_ADAPTERKEY;
+            queryArgs.QueryFlags.TranslatePath = TRUE;
+            queryArgs.ValueType = REG_SZ;
+	    result = MultiByteToWideChar(
+                	CP_ACP,
+                	0,
+                	cszOpenCLRegKeyName,
+                	szOpenCLRegKeyName,
+                	queryArgs.ValueName,
+                	ARRAYSIZE(queryArgs.ValueName));
```
ofirc: don't we need an strlen here? ARRAYSIZE I assume is a compile-time construct, I wonder if the actual size could be less then the fixed compile time size.

ofirc:
See also other inline comments in the code.

loader/windows/icd_windows_dxgk.c
nikhiljnv: Shouldn't this be taken care by outer sizeof(cszOpenCLRegKeyName[0])? Multiplication by sizeof(cszOpenCLRegKeyName[0]) applies to 1 as well.

loader/windows/icd_windows.c
nikhiljnv: Keeping it to be on the safer side.

loader/windows/icd_windows.c
nikhiljnv: > 
> 
> Looks better, thanks for the fixes.
> I'm still curious about when do we release the resources (i.e. the dynamically allocated list that gets incremented by 2 in its size), I haven't seen the free for it (i.e. DllMain / callback from the OS on library unload.

The older list gets freed up as soon as new list with twice the size is allocated in adapterAdd.
The final list gets freed up in free after call to adapterFree once all adapters are enumerated.

loader/windows/icd_windows_dxgk.c
nikhiljnv: > 
> 
> Indentation inconsistency.

Fixed.

loader/windows/icd_windows_dxgk.c
nikhiljnv: > 
> 
> ditto

Fixed.

loader/windows/icd_windows.c
nikhiljnv: > 
> 
> could just use this way instead? https://docs.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-createdxgifactory

Saleel,
Are you suggesting to call the API directly without calling GetProcAddress?
The API may not be available on all Windows versions. So, it would be good to go through GetProcAddress.

loader/windows/icd_windows.c
```diff
@@ -23,8 +23,75 @@
...
+    if (pWinAdapterEnd == pWinAdapterCapacity)
+    {
+        size_t oldCapacity = pWinAdapterCapacity - pWinAdapterBegin;
+        size_t newCapacity = oldCapacity;
+        if (0 == newCapacity)
+        {
+            newCapacity = 1;
+        }
+	else 
+            newCapacity *= 2;
```
ofirc: We should probably protect against unsigned integer overflow here:
else
{
     KHR_ICD_ASSERT(newCapacity < UINT_MAX / 2);
     newCapacitiy *= 2;
}

loader/windows/icd_windows_dxgk.c
```diff
@@ -83,47 +88,48 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
+            result = MultiByteToWideChar(
+                CP_ACP,
+                0,
+                cszOpenCLRegKeyName,
+                szOpenCLRegKeyName,
+                queryArgs.ValueName,
+                ARRAYSIZE(queryArgs.ValueName));
+	    if (!result)
+	    {
+            	KHR_ICD_TRACE("MultiByteToWideChar status != SUCCESS\n");
```
ofirc: Inconsistent indentation

loader/windows/icd_windows_dxgk.c
```diff
@@ -83,47 +88,48 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
             if (NT_SUCCESS(Status) && pQueryArgs->Status == D3DDDI_QUERYREGISTRY_STATUS_BUFFER_OVERFLOW)
             {
-                ULONG QueryBufferSize = sizeof(D3DDDI_QUERYREGISTRY_INFO) + QueryArgs.OutputValueSize;
-                pQueryBuffer = (D3DDDI_QUERYREGISTRY_INFO*)malloc(QueryBufferSize);
-                memcpy(pQueryBuffer, &QueryArgs, sizeof(D3DDDI_QUERYREGISTRY_INFO));
-                QueryAdapterInfo.pPrivateDriverData = pQueryBuffer;
-                QueryAdapterInfo.PrivateDriverDataSize = QueryBufferSize;
-                Status = D3DKMTQueryAdapterInfo(&QueryAdapterInfo);
+                ULONG queryBufferSize = sizeof(D3DDDI_QUERYREGISTRY_INFO) + queryArgs.OutputValueSize;
+                pQueryBuffer = (D3DDDI_QUERYREGISTRY_INFO*)malloc(queryBufferSize);
```
ofirc: We need to check the result of malloc.

loader/windows/icd_windows.c
ofirc: This is not safer, this is just redundant:
https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c

loader/windows/icd_windows.c
ofirc: In which file? I don't see the call to free all adapters after adapterFree.

nikhiljnv: > 
> 
> In which file? I don't see the call to free all adapters after adapterFree.

It's is the same file icd_windows.c It's just a single free on pWinAdapterBegin as this is single malloced memory.

loader/windows/icd_windows_dxgk.c
nikhiljnv: ValueName isn't reallocated after queryArgs is declared. Agree that it is better to use strlen in general, but shouldn't be strictly needed here.

loader/windows/icd_windows.c
nikhiljnv: > 
> 
> We should probably protect against unsigned integer overflow here:
> else
> {
> KHR_ICD_ASSERT(newCapacity < UINT_MAX / 2);
> newCapacitiy *= 2;
> }

Fixed.

nikhiljnv:
@ofirc @saleelk 
Are we good to go here? Please, share feedback if any by tomorrow EOD.

saleelk:
LGTM, thanks.

loader/windows/icd_windows.c
saleelk: Yea, though checking free isnt really needed, I regularly see crashes elsewhere in the driver when not checking if a pointer exists. Though this is probably due to bad practices that are followed elsewhere and tricked down to the crash spot... so it has become a trend to check

nikhiljnv:
@bashbaug ,
Can we merge this one?

bashbaug:
I have these changes rebased and ready to go.  Let's approve and merge in tomorrow's teleconference.

bashbaug:
Rebased and pushed to master as 978b4b3a29a3aebc86ce9315d5c5963e88722d03.

PR #86, opened 2019-07-30
Install to standard locations using GNUInstallDirs

candrews:
Different distributions may specify different paths; using GNUInstallDirs ensures that everything installed to the correct location for all systems. For example, amd64 Debian uses lib64 as opposed to lib, so hardcoding lib would be wrong for Debian.


bashbaug:
LGTM.  Waiting another day or so for other OpenCL WG reviewers.  Thanks!

PR #88, opened 2019-08-30
added checks for high integrity for Windows secure_getenv

bashbaug:
This PR adds an analogy to secure_getenv for Windows by checking the "integrity level" of the executing process and returning NULL when the process is executing with elevated privileges.


nikhiljnv:
LGTM.

bashbaug:
Merging as per OpenCL WG discussion.

PR #89, opened 2019-11-08
Allow OpenCL library name to be changed from the outside

eldruin:
Renaming the libraries is sometimes used as to avoid interference with the library versions already available on the target systems by giving them a prefix.
This PR allows users to specify, for example:
```
-DOPENCL_OUTPUT_NAME:STRING=MYORG_OpenCL
```


eldruin:
I will need to ask my employer about the CLA even though this change is really small.
Would you accept the change provided the CLA was signed?
(I would like to avoid overhead on our side if not necessary)

PR #90, opened 2019-11-18
Remove link-time dependency on WDK 

nikhiljnv:
Remove link-time dependency on WDK 

Recent changes for using QueryAdapter and DXGI APIs
for enumerating platforms on Windows added a dependency on
WDK for linking OpenCL driver to Windows specific
libraries.

Querying function address using GetProcAddress for new QueryAdapterInfo
function so that there is no need to link to WDK while
building OpenCL ICD loader.

Also, removing the OPENCL_ICD_LOADER_REQUIRE_WDK guard
needed because of link-time dependency.

While at it, fixing variable names to use camelCase.
Also, use realloc instead of malloc and copy while resizing
memory for adapters.


nikhiljnv:
@bashbaug ,
Can you take a look and see if anything is left here?

CMakeLists.txt
```diff
@@ -80,26 +80,23 @@ set_target_properties (OpenCL PROPERTIES VERSION "1.2" SOVERSION "1")
...
     target_link_libraries (OpenCL cfgmgr32.lib)
 
-    option (OPENCL_ICD_LOADER_REQUIRE_WDK "Build with D3DKMT support, which requires the Windows WDK." ON)
-    if (OPENCL_ICD_LOADER_REQUIRE_WDK)
-        if(DEFINED ENV{WDKContentRoot})
-            file(GLOB D3DKMT_HEADER "$ENV{WDKContentRoot}/Include/*/km/d3dkmthk.h")
-        else()
-            file(GLOB D3DKMT_HEADER "$ENV{HOMEDRIVE}/Program Files */Windows Kits/10/Include/*/km/d3dkmthk.h")
-        endif()
+    if(DEFINED ENV{WDKContentRoot})
```
bashbaug: It looks like the WDK is still required after these changes, specifically for the WDK headers, which is unfortunate.

Can the OpenCL ICD loader use its own header instead, similar to the Vulkan loader?

* https://github.com/KhronosGroup/Vulkan-Loader/commit/7557a5753409f89d78eaaf23367cecc60574b3be

I think this would break all dependencies on the WDK?

CMakeLists.txt
```diff
@@ -80,26 +80,23 @@ set_target_properties (OpenCL PROPERTIES VERSION "1.2" SOVERSION "1")
...
+    	list(GET D3DKMT_HEADER -1 LATEST_D3DKMT_HEADER)
+        get_filename_component(WDK_DIRECTORY ${LATEST_D3DKMT_HEADER} DIRECTORY)
+        get_filename_component(WDK_DIRECTORY ${WDK_DIRECTORY} DIRECTORY)
+        message(STATUS "Found the Windows WDK in: ${WDK_DIRECTORY}")
+        target_compile_definitions(OpenCL PRIVATE OPENCL_ICD_LOADER_REQUIRE_WDK)
+        target_include_directories(OpenCL PRIVATE ${WDK_DIRECTORY}/um)
+        target_include_directories(OpenCL PRIVATE ${WDK_DIRECTORY}/km)
+        target_include_directories(OpenCL PRIVATE ${WDK_DIRECTORY}/shared)
+    else()
+        message(FATAL_ERROR "The Windows WDK was not found. Consider disabling OPENCL_ICD_LOADER_REQUIRE_WDK. Aborting.")
```
bashbaug: I'd still prefer to break any dependency on the WDK, but if we cannot, this message is now incorrect and needs to be updated.

loader/windows/icd_windows.c
```diff
@@ -58,16 +58,11 @@ BOOL adapterAdd(const char* szName, LUID luid)
             newCapacity *= 2;
         }
 
-        WinAdapter* pNewBegin = malloc(newCapacity * sizeof(*pWinAdapterBegin));
+        WinAdapter* pNewBegin = realloc(newCapacity * sizeof(*pWinAdapterBegin));
```
bashbaug: Can you check that this is the correct use of `realloc`?  I think this is incorrect and it's causing CI builds to fail.

CMakeLists.txt
```diff
@@ -80,26 +80,23 @@ set_target_properties (OpenCL PROPERTIES VERSION "1.2" SOVERSION "1")
...
-            target_include_directories(OpenCL PRIVATE ${WDK_DIRECTORY}/shared)
-        else()
-            message(FATAL_ERROR "The Windows WDK was not found. Consider disabling OPENCL_ICD_LOADER_REQUIRE_WDK. Aborting.")
-        endif()
+    if(D3DKMT_HEADER)
+    	list(GET D3DKMT_HEADER -1 LATEST_D3DKMT_HEADER)
+        get_filename_component(WDK_DIRECTORY ${LATEST_D3DKMT_HEADER} DIRECTORY)
+        get_filename_component(WDK_DIRECTORY ${WDK_DIRECTORY} DIRECTORY)
+        message(STATUS "Found the Windows WDK in: ${WDK_DIRECTORY}")
+        target_compile_definitions(OpenCL PRIVATE OPENCL_ICD_LOADER_REQUIRE_WDK)
```
bashbaug: Assuming we can break all dependencies on the WDK we should be able to get rid of this `OPENCL_ICD_LOADER_REQUIRE_WDK` define.

loader/windows/icd_windows.c
bashbaug: This is fixed and it looks like CI builds are passing now.

bashbaug:
Hi @nikhiljnv , one of the changes I requested has been addressed, but I think the other comments still apply.

CMakeLists.txt
nikhiljnv: I have tentatively update comment. I will continue to work on breaking header dependencies Vulkan way. If I can get it done this week, we can merge it as part of this change. Otherwise, I prefer merging this change as is and then break header dependencies separately.

bashbaug:
Hi @nikhiljnv , now that #102 is merged, is there anything more to add in this PR?  Thanks!

nikhiljnv:
Closing this in favor of #102 which is already merged.

PR #92, opened 2019-12-09
Update and modernize some aspects of CMake

kbenzie:
Ease integration with projects targeting more recent CMake versions:

* Update minimum required CMake version to 3.1.
* Specify a version `1.2` to the `project` CMake command, this silences
  developer warnings when the OpenCL-ICD-Loader is integrated in another
  project using `add_subdirectory`.
* Specify include directory on the `OpenCL` target, this removes the
  need to set the OpenCL-Headers include directory globally (old CMake)
  or set them manually per target which links the `OpenCL` target
  library.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,6 +15,6 @@ before_build:
   - cmake -DOPENCL_ICD_LOADER_HEADERS_DIR=inc/OpenCL-Headers -H. -Bbuild -A%PLATFORM%
 
 build:
-  project: build\OPENCL_ICD_LOADER.sln
+  project: build\OpenCL-ICD-Loader.sln
   parallel: true
   verbosity: normal
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
-cmake_minimum_required (VERSION 2.8.11)
+cmake_minimum_required (VERSION 3.1)
 
-project (OPENCL_ICD_LOADER)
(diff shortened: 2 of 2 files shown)


MathiasMagnus:
- I don't have any objections against upping the minimum version to 3.1.
- Having a project version is also desirable. Perhaps not in this PR, but it should follow the version which the ICD exposes. In current master there is an option which imports 3.0 entry points, in which case the ICD version should follow.
  - I do think however that importing a project via `add_subdirectory` is bad practice. (Some tools tolerate / workaround encountering multiple `project()` invocations, but top-level CMakeLists.txt files should be processed using `ExternalProject_Add()` [as per the docs](https://cmake.org/cmake/help/latest/module/ExternalProject.html).
- That seems to be appropriate, indeed.

kbenzie:
> * Having a project version is also desirable. Perhaps not in this PR, but it should follow the version which the ICD exposes. In current master there is an option which imports 3.0 entry points, in which case the ICD version should follow.

I'm unsure if I need to take action for this point.

I can see the logic of using the most recent version supported by the ICD-Loader. Should this be spun out into a separate ticket?

The reason for setting this to `1.2` is that is the shared object version used for the `OpenCL` target, should this also reflect the latest supported version?

MathiasMagnus:
Yes, we can always merge and refine according to the other PRs when they get ironed out.

kbenzie:
Seems like everyone is happy with this, would it be possible for someone to press the accept button?

PR #85, opened 2019-07-19
Allow REG_MULTI_SZ as OpenCLDriverName type

dooglz:
Fix for #84

Issue #84, opened 2019-07-19
AMD opencl uses REG_MULTI_SZ  for OpenCLDriverName


dooglz:
I Have a windows system with both Nvidia and AMD gpus, both individual vendor opencl driver work when used directly. Using this ICD the amd driver fails to be detected because AMD have used 
**REG_MULTI_SZ** rather than **REG_SZ** for the OpenCLDriverName.

I know this is technically AMD's fault as they are going against the docs here
https://github.com/KhronosGroup/OpenCL-Docs/blob/master/ext/cl_khr_icd.txt#L71

If the maintainers want to allow this small deviation from the spec: See pull request for quick fix. #85

saleelk:
The issue has been fixed in AMD trunk and new drivers should ship with REG_SZ

b-sumner:
This can be closed since it is no longer necessary.

bashbaug:
Closing as per the [comment above](https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/84#issuecomment-520896371).

PR #94, opened 2020-02-07
Fix to allow Windows 10 (WDDM 2.4) ICD detection to detect AMD ICD drivers

johngladp:
The ICD loader doesn't detect AMD ICD drivers on Windows 10 (WDDM 2.4).  This was due to AMD driver INF writing out the OpenCLDriverName registry key using the REG_MULTI_SZ value type for their ICD keys; the loader assumes a REG_SZ key type; Intel INF files use the REG_SZ value type.  

Using the AMD built and provided OpenCL.dll (build date 7th Dec 2018) results in only the AMD ICD being detected.

When the D3DKMTQueryAdapterInfo operation fails (it's returning STATUS_INVALID_PARAMETER) we can unconditionally re-try the query operation with the alternate value type resulting in possible success.  

This change allows the ICD loader to detect both AMD and Intel ICD drivers.


johngladp:
Issue #84 appears to be the cause of this and whilst AMD are addressing it that doesn’t help for existing systems with latest drivers that don’t include the AMD driver INF fix.

loader/windows/icd_windows_dxgk.c
```diff
@@ -118,9 +118,15 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
             Status = D3DKMTQueryAdapterInfo(&queryAdapterInfo);
             if (!NT_SUCCESS(Status))
             {
-                // Continue trying to get as much info on each adapter as possible.
-                // It's too late to return FALSE and claim WDDM2_4 enumeration is not available here.
-                continue;
+                // Try a different value type.  Some vendors write the key as a multi-string type.
+                queryArgs.ValueType = REG_MULTI_SZ;
+                Status = D3DKMTQueryAdapterInfo(&queryAdapterInfo);
```
bashbaug: This seems reasonable.  A few administrative topics:

1. I think we should bump the ICD loader version number with this change, so ICD loaders with this fix are considered newer and overwrite ICD loaders without the change.
1. Do we want to consider this a workaround for drivers/installers that do not follow the cl_khr_icd spec, or do we want to update the spec to say that these keys may either be REG_SZ or REG_MULTI_SZ?
1. Do we want to apply a similar change to the HKR codepath?  I believe it is currently only checking for REG_SZ.
1. Regardless, I'd suggest adding some KHR_ICD_TRACE messages to trace when the ICD loader switches from a REG_SZ key to a REG_MULTI_SZ key.

Thanks!

johngladp: Hi,

I've incorporated points 1, 3 & 4 into a second commit.  I have tested the DXGK and HKR code paths briefly on my local system and both enumerate all platforms as expected with these changes.

Regarding point 2.  I would suggest we consider this a workaround.  Comments in the closed pull request #85 suggest AMD will eventually abide by the specification but there are drivers in the field necessitating this workaround.

Thanks for the code review!
John A.

johngladp:
I am happy to wait.
Thanks again for the code review!
John A.

ofirc:
I'll review it over the next few days.
Quick question, why not align across the board on what type of reg key to use?
@bashbaug isn't there a strict spec for drivers for this?

bashbaug:
@ofirc, yes, there is a spec, and drivers should be using `REG_SZ`, see Section 2.4 in the [cl_khr_icd spec](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Ext.html#cl_khr_icd-opencl).  This change is a workaround for older drivers that did not properly implement the spec and used `REG_MULTI_SZ` instead, which otherwise do not enumerate properly when used with newer ICD loaders.

loader/windows/icd_windows_hkr.c
```diff
@@ -131,8 +131,15 @@ static bool ReadOpenCLKey(DEVINST dnDevNode)
 
         if (REG_SZ != dwLibraryNameType)
```
ofirc: Cosmetic: we could simplify the code below (and thereby decrease the indentation level) by:
if (REG_SZ != dwLibraryNameType && REG_MULTI_SZ == dwLibraryNameType)
{
    KHR_ICD_TRACE("Accepting multi-string registry key type\n");
}
else
{
    KHR_ICD_TRACE("Unexpected registry entry 0x%x! continuing\n", dwLibraryNameType);
}

loader/windows/icd_windows_dxgk.c
ofirc: Re point 2 (workaround), I'm quoting Ben's response on this matter:
"drivers should be using REG_SZ, see Section 2.4 in the [cl_khr_icd spec](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Ext.html#cl_khr_icd-opencl). This change is a workaround for older drivers that did not properly implement the spec and used REG_MULTI_SZ instead, which otherwise do not enumerate properly when used with newer ICD loaders."

Something is not clear to me.
The ICD Loader is (usually) shipped and coupled with the accompanying driver (say Graphics Driver or independent OpenCL Runtime/Platform implementation), so whenever this workaround is applied it means that you shipped a new driver bundle.
Why not fix the root cause and adapt the INF file that comes with the Graphics Driver to use REG_MULTI_SZ?
You're already shipping a new ICD Loader, why not ship a "fixed"/conforming driver while you're at it?

In other words, under what environments does this bug reproduce? hybrid? (e.g. Intel CPU + AMD GPU)

loader/windows/icd_windows_hkr.c
```diff
@@ -131,8 +131,15 @@ static bool ReadOpenCLKey(DEVINST dnDevNode)
...
         {
-            KHR_ICD_TRACE("Unexpected registry entry 0x%x! continuing\n", dwLibraryNameType);
-            goto out;
+            if (REG_MULTI_SZ == dwLibraryNameType)
+            {
+               KHR_ICD_TRACE("Accepting multi-string registry key type\n");
```
ofirc: Ditto.

loader/windows/icd_windows_hkr.c
```diff
@@ -131,8 +131,15 @@ static bool ReadOpenCLKey(DEVINST dnDevNode)
...
+            }
+            else
+            {
+               KHR_ICD_TRACE("Unexpected registry entry 0x%x! continuing\n", dwLibraryNameType);
```
ofirc: Ditto.

loader/windows/icd_windows_dxgk.c
```diff
@@ -121,7 +121,11 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
                 // Try a different value type.  Some vendors write the key as a multi-string type.
                 queryArgs.ValueType = REG_MULTI_SZ;
                 Status = D3DKMTQueryAdapterInfo(&queryAdapterInfo);
-                if (!NT_SUCCESS(Status))
+                if (NT_SUCCESS(Status))
+                {
+                   KHR_ICD_TRACE("Accepting multi-string registry key type\n");
```
ofirc: Inconsistent indentations, use 4 spaces.

loader/windows/icd_windows_dxgk.c
```diff
@@ -118,9 +118,15 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
-                // Continue trying to get as much info on each adapter as possible.
-                // It's too late to return FALSE and claim WDDM2_4 enumeration is not available here.
-                continue;
+                // Try a different value type.  Some vendors write the key as a multi-string type.
+                queryArgs.ValueType = REG_MULTI_SZ;
+                Status = D3DKMTQueryAdapterInfo(&queryAdapterInfo);
+                if (!NT_SUCCESS(Status))
+                {
+                   // Continue trying to get as much info on each adapter as possible.
+                   // It's too late to return FALSE and claim WDDM2_4 enumeration is not available here.
```
ofirc: early or late :)?

loader/windows/icd_windows_dxgk.c
```diff
@@ -118,9 +118,15 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
...
             {
-                // Continue trying to get as much info on each adapter as possible.
-                // It's too late to return FALSE and claim WDDM2_4 enumeration is not available here.
-                continue;
+                // Try a different value type.  Some vendors write the key as a multi-string type.
+                queryArgs.ValueType = REG_MULTI_SZ;
+                Status = D3DKMTQueryAdapterInfo(&queryAdapterInfo);
+                if (!NT_SUCCESS(Status))
+                {
+                   // Continue trying to get as much info on each adapter as possible.
```
ofirc: Indentation, use 4 spaces.

loader/windows/icd_windows_dxgk.c
johngladp: This is reproducible on a system with Intel CPU Opencl, Intel GPU Opencl and AMD GPU Opencl (Vega-m), all latest drivers. The AMD loader enumerates only the AMD platform devices, the Intel loader enumerates only the Intel platform devices. I would suggest it is correct to be defensive and enumerate all platforms, including those that have deviated from the spec.

loader/windows/icd_windows_dxgk.c
johngladp: I will incorporate all of these requested changes and push the changes on Monday (17th).

loader/windows/icd_windows_hkr.c
johngladp: This logic isn't equivalent. It would catch the REG_SZ key type as an invalid registry entry.  I'll keep the block as is, it makes it clear to the reader than the REG_MULTI_SZ is considered to be an anomaly.

loader/windows/icd_windows_dxgk.c
ofirc: Then what good is the spec for :)? is it a recommendation for vendors or something to adhere to?
It's OK to put workaround for the greater good, or at least temporarily, but IMHO it should be discouraged.

loader/windows/icd_windows_hkr.c
ofirc: You're right, my bad.

ofirc:
The changes looks good to me, thanks for the patch.
The same way OpenCL conformance tests verify the spec (API + device side), we should be able to minimize deviance from the spec on the cl_khr_icd extension as well, that's my 2 cents.
A TODO/FIXME note would clarify this particular deviance.

loader/windows/icd_windows_dxgk.c
ofirc: On second thought, we are required (or at least expected) to be backwards compatible in some sense. So if we changed from REG_SZ to REG_MULTI_SZ at some point in time, it makes sense to accept both as the loader should be able to enumerate older conforming platform implementations.

bashbaug:
Hi @johngladp , thank you very much for your contribution!  Merging these changes as per approval in the February 25th working group teleconference.

PR #103, opened 2020-04-09
Add support for OpenCLOn12 ICD

jenatali:
The OpenCLOn12 project ([announced here](https://devblogs.microsoft.com/directx/in-the-works-opencl-and-opengl-mapping-layers-to-directx/)) doesn't look like a normal ICD. On Windows, ICDs are enumerated in three ways:
1. Explicitly belonging to a specific GPU driver installation (GPU-specific registry entries).
2. Added to a global set of ICDs installed on the system in the registry.
3. Set in the environment.

The OpenCLOn12 ICD won't be tied to any given hardware, and its distribution mechanism doesn't allow to modify the registry. This ICD will be shipping inside an app package, delivered by the Windows store. We'd like to add explicit logic to the ICD loader to look for this package, and if installed, load the ICD out of it.

Since this ICD doesn't have an adapter LUID associated with it, its platform will be relegated to the end of the platform list. This behavior is probably for the best.


bashbaug:
@nikhiljnv  @b-sumner @saleelk did you want to review these changes?  If so, can you please provide feedback by EOW?

saleelk:
LGTM

nikhiljnv:
Looks good to me !!

MathiasMagnus:
> This ICD will be shipping inside an app package, delivered by the Windows store. 

+1 for the distribution method. Eagerly awaiting GA (or a Store link 😉).

loader/windows/icd_windows_apppackage.cpp
```diff
@@ -36,7 +36,12 @@ using namespace Microsoft::WRL::Wrappers;
 
 extern "C" bool khrIcdOsVendorsEnumerateAppPackage()
 {
-    if (FAILED(Windows::Foundation::Initialize()))
+    HRESULT hrInit = Windows::Foundation::Initialize(RO_INIT_MULTITHREADED);
+    if (hrInit == RPC_E_CHANGED_MODE)
+    {
+        hrInit = Windows::Foundation::Initialize(RO_INIT_SINGLETHREADED);
```
cubanismo, line 54: Does this only happen in cases where the application itself is expected to be single threaded?  I don't really understand the implications of falling back, or in what cases it would happen, but if the assumption above is true, seems good.

loader/windows/icd_windows_apppackage.cpp
```diff
@@ -46,6 +58,10 @@ extern "C" bool khrIcdOsVendorsEnumerateAppPackage()
         KHR_ICD_TRACE("Failed to init WinRT\n");
         return false;
     }
+    auto Cleanup = MakeScopeExit([]()
+        {
+            Windows::Foundation::Uninitialize();
```
cubanismo, line 63: Should this be included in the init change?  I'm not clear if you sent separate commits just for ease of review (Thank you!) and plan to squash them, or if they'll go in separately.  It seems like init and unint should come as one piece though.

loader/windows/icd_windows.c
```diff
@@ -256,7 +256,12 @@ void khrIcdOsVendorsEnumerateOnce()
 // dynamically load a library.  returns NULL on failure
 void *khrIcdOsLibraryLoad(const char *libraryName)
 {
-    return (void *)LoadLibraryA(libraryName);
+    HMODULE hTemp = LoadLibraryExA(libraryName, NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
```
cubanismo, line 259: Glad this encapsulated so well :-)

loader/windows/icd_windows_apppackage.cpp
```diff
@@ -129,7 +129,7 @@ extern "C" bool khrIcdOsVendorsEnumerateAppPackage()
         }
 
         HString path;
-        if (FAILED(item->get_Path(path.ReleaseAndGetAddressOf())))
+        if (FAILED(item->get_Path(path.GetAddressOf())))
```
cubanismo, line 132: I assume this should ultimately be squashed with the first commit as well?

loader/windows/icd_windows_apppackage.cpp
jenatali: Yeah, if the app has already initialized COM or WinRT as singlethreaded on the thread where the loader is being initialized, then we have to use singlethreaded as well. In all other cases (already initialized as multi-threaded, or uninitialized), we'll use multithreaded.

loader/windows/icd_windows_apppackage.cpp
jenatali: I was planning to squash them together. I can go ahead and squash those together ahead-of-time, in case you'd prefer to keep these separate after merging.

loader/windows/icd_windows_apppackage.cpp
jenatali: Yes, this is a fixup to support older SDKs. I had no idea that function wasn't there in older SDKs.

bashbaug:
Thanks everyone!  Merging.

neiltrevett:
Thanks everyone!
@jenatali - we will use the weekly OpenCL calls to push reviews forward in a timely way - hopefully this one didn't hold you back too badly. If you need to alert the group to speed up a review of time sensitive merges please shout on the issue and/or ICD list and we will do our best to respond.

PR #108, opened 2020-04-24
Add support for OpenCL 3.0 Provisional

alycm:
The new entry points will only be enabled if `CL_EXPERIMENTAL` is manually set.

Note that even a non experimental libOpenCL.so ends up with an `OPENCL_EXPERIMENTAL` absolute symbol on Linux builds.  I don't think that that is a problem though, as it is not referenced.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,14 @@ find_package (Threads REQUIRED)
 # advance. Use it with discretion.
 option (BUILD_SHARED_LIBS "Build shared libs" ON)
 
+# This option enables support for OpenCL 3.0 Provisional in the ICD loader.  It
+# is currently off by default while the specification is provisional, as it may
+# change.
+# NOTE: If setting this to ON on Windows you likely need to uncomment the last
+# two entries in loader/windows/OpenCL.def.  When this option is deleted those
+# entries should be permanently uncommented.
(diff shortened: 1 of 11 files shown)


alycm:
I obviously skipped a lot of untested entry points in the PR, I've filed an issue about actually implementing those tests: https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/109.

alycm:
I have updated this to OpenCL 3.0 provisional.  This is still off by default, the OPENCL_3.0 is present in Linux builds regardless, but this seems harmless.

Windows builds with 3.0 enabled may require adding two symbols to the export map, this could be resolved by having CMake generate that file, but I have no done that currently.

alycm:
My thought was to use cmake `configure_file` or an equivalent mechanism, I'll take a look soon.

alycm:
I've updated this PR to use `configure_file` to generate an `OpenCL.def` or `icd_exports.map` that will only include 3.0 symbols when `ENABLE_OPENCL30_PROVISIONAL` is set.  No manual intervention required.

Testing Linux both ways locally, Windows in progress.

alycm:
Also now tested locally on Windows.

bashbaug:
Thanks!  I've verified that this works with no manual intervention also.  Merging!

PR #110, opened 2020-04-30
add an option to disable support for OpenCLOn12

bashbaug:
This PR adds a CMake option to disable support for the OpenCLOn12 ICD, which reduces the build dependencies for the OpenCL ICD loader.  This option should NOT be used for building an OpenCL ICD loader DLL for distribution, but it can be used to generate a DLL import library to link against.

See #107.

I am marking this PR WIP until we can verify it solves #107 and there is no better solution.


jenatali:
I've got a MinGW installation, and I see a libruntimeobject.a present, so that should be fine.

bashbaug:
Since this change fixes #107 I think it's something we should pursue, at least as a short-term solution.  I've removed WIP and I think it's ready to go.

PR #113, opened 2020-07-24
add clSetContextDestructorCallback

### LINKED ISSUE #112: Add clSetContextDestructorCallback for OpenCL 3.0
The 3.0 specification added a new API entry point, this should be added to the ICD loader.  Needs header support first, see https://github.com/KhronosGroup/OpenCL-Headers/issues/98.

https://github.com/KhronosGroup/OpenCL-Docs/pull/346

bashbaug:
Adds the recently added `clSetContextDestructorCallback` API function to the ICD loader.

See: https://github.com/KhronosGroup/OpenCL-Docs/issues/187 https://github.com/KhronosGroup/OpenCL-Docs/pull/346

Fixes #112.

### DIFF
--- a/test/driver_stub/cl.c
+++ b/test/driver_stub/cl.c
@@ -368,10 +368,10 @@ clSetContextDestructorCallback(cl_context         context,
                       context,
                       pfn_notify,
                       user_data);
-    pfn_notify(context, NULL);
+    pfn_notify(context, user_data);
     test_icd_stub_log("setcontextdestructor_callback(%p, %p)\n",
                context,
-               NULL);
+               user_data);
 
     test_icd_stub_log("Value returned: %d\n", return_value);
     return return_value;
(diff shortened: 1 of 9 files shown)


bashbaug:
Note, ideally we would merge the header changes first (https://github.com/KhronosGroup/OpenCL-Headers/pull/101), but because OpenCL 3.0 is disabled by default we could merge these changes ASAP.  The updated header files are needed when OpenCL 3.0 is enabled, though.

test/driver_stub/cl.c
```diff
@@ -355,6 +355,30 @@ clGetContextInfo(cl_context         context,
...
+{
+    cl_int return_value = CL_OUT_OF_RESOURCES;
+    test_icd_stub_log("clSetContextDestructorCallback(%p, %p, %p)\n",
+                      context,
+                      pfn_notify,
+                      user_data);
+    pfn_notify(context, NULL);
+    test_icd_stub_log("setcontextdestructor_callback(%p, %p)\n",
+               context,
+               NULL);
```
alycm, line 374:
```suggestion
    pfn_notify(context, user_data);
    test_icd_stub_log("setcontextdestructor_callback(%p, %p)\n",
               context,
               user_data);
```

I didn't actually try this out.

It doesn't make any difference to the tests, so not too important whether the stub does this or not.

bashbaug: Interesting - it looks like all of the tests pass `NULL` unconditionally to the callbacks vs. `user_data`.

I've fixed this for both `clSetMemObjectDestructorCallback` and `clSetContextDestructorCallback` with the change suggested above. I didn't touch the other callbacks for now though, and if we want to fix these I think we should do it as a follow-on PR.

Thanks!

alycm: Looks good, thanks!

PR #114, opened 2020-08-05
Generate the ICD Loader Dispatch Functions

bashbaug:
This PR adds a script to generate the ICD loader dispatch functions from the cl.xml machine readable grammar vs. implementing them "manually".  The dispatch functions are generated by parsing the XML file into a Python dictionary, then generating the file using a [Mako](https://www.makotemplates.org/) template.

The generated functions are mostly the same as the manually written functions, but there are some key differences.  Note, I collected these differences by running clang-format on the manual functions and the generated functions, reshuffling the manual functions to they are in the same order as the generated functions, then diff'ing the two files:

1. There are fewer calls to `khrIcdInitialize`, see #73.
2. The generated `clSVMFree` only checks if the `context` argument is `NULL` and does not return early if the `svm_pointer` argument is `NULL`.  I think I can add the extra check if desired, but it's not clear to me that it is required.
3. The generated `clSetDefaultDeviceCommandQueue` only checks if the `context` argument is `NULL` and does not return early if the `device` argument is `NULL`.  I think I can add this extra check if desired, as well.
4. The generated `clGetKernelSubGroupInfo` only checks if the `kernel` argument is `NULL` and does not return early if the `device` argument is `NULL`.  I can add this extra check if desired, as well, though do note that there is no `device` check for `clGetKernelWorkGroupInfo`.
5. There are minor cosmetic differences in the names of some arguments and for some of the API suffixes, but these differences should be purely cosmetic.

The ICD loader tests are passing after these changes.

If these changes look fundamentally good, I'll write up a short README describing how to install any dependencies and run the generator script.  I can also bump the file version, if desired, so we can distinguish between the last manual ICD loader and the first generated ICD loader.

Thanks!


Kerilk:
1. I wasn't there when you opened https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/73, but what is the state of the discussion regarding NULL platform support? If the NULL platform defaults to another platform, then the initializer must be called before any function that takes a platform, so the default platform can be determined. (ocl-icd does that, and clinfo checks for the behavior). We could imagine implementing that as a layer though, which would always call clGetPlatformIDs first thus solving the issue.
2. This just avoids a function call. We could formalize those additions by registering prologues and epilogues for each function that needs special processing before and after dispatch. But I am unsure any check not related to dispatch should be made in the loader.
3. see 2.
4. see 2.

Aside from the discussion around 1. the loader seems fine to me as you generate it.

bashbaug:
> 1. what is the state of the discussion regarding NULL platform support? If the NULL platform defaults to another platform, then the initializer must be called before any function that takes a platform, so the default platform can be determined. (ocl-icd does that, and clinfo checks for the behavior). We could imagine implementing that as a layer though, which would always call clGetPlatformIDs first thus solving the issue.

Good question.  I've taken another look at the deltas with this in mind.  Here is what I see, for each affected API:

**clGetPlatformInfo**: 
* Spec says "If platform is NULL, the behavior is implementation-defined."
* The previous "manual" implementation of this function called `khrIcdInitialize` but didn't choose a "default" platform if the incoming platform was `NULL`.  So, there should be no functional differences switching to the generated implementation.

**clGetDeviceIDs**:
* Spec says "If platform is NULL, the behavior is implementation-defined."
* Similarly, the previous "manual" implementation of this function called `khrIcdInitialize` but didn't choose a "default" platform if the incoming platform was `NULL`.  So, there should be no functional differences switching to the generated implementation.

**clCreateContext**:
* Spec says "properties can be NULL in which case the platform that is selected is implementation-defined."
* There should be no change in behavior by removing `khrIcdInitialize` from this function because it calls into the ICD through a device ID and does not parse the properties.

**clUnloadPlatformCompiler**
* Spec does not describe any behavior if the platform is `NULL`, only that it may return `CL_INVALID_PLATFORM` "if platform is not a valid platform."
* The previous "manual" implementation called `khrIcdInitialize`, but both the manual implementation and the generated implementation returns `CL_INVALID_PLATFORM` for a `NULL` platform, so there should be no functional differences switching to the generated implementation.

So, while we could choose different implementation-defined behavior for these APIs, I don't think there will be any functional differences between the manual versions and their generated equivalents.

Note that I didn't change **clGetExtensionFunctionAddressForPlatform**, but given that it returns a `NULL` function pointer when the incoming platform is `NULL` vs. choosing a default platform I think its call to `khrIcdInitialize` is unnecessary and can be removed, also.

Kerilk:
This behavior has always seemed incoherent to me with the content of the extension:
https://github.com/KhronosGroup/OpenCL-Docs/blob/master/ext/cl_khr_icd.asciidoc#L334-L344
which state:
> 3. How will the ICD handle a NULL cl_platform_id?
>
>RESOLVED: The ICD will by default choose the first enumerated platform as the NULL platform. The user can override this default by setting an environment variable OPENCL_ICD_DEFAULT_PLATFORM to the desired platform index. The API calls that deal with platforms will return CL_INVALID_PLATFORM if the index is not between zero and (number of platforms - 1), both inclusive.
I know the wording is strange but always believed ICD was ICD loader in this paragraph. 

So my understanding is that this should not be implementation defined, and that the previous behavior of the loader was not correct.
In the case of `clCreateContext` it could be argued that if the user provided a platform it should be used for dispatch. If the user provided a NULL platform, maybe it should be considered the default platform (I noticed this last case is not correctly handled in ocl-icd).

bashbaug:
I've added a README describing how the file generation works and I've incremented the file version.  I think this is ready to go now.  Thanks!

scripts/gen_loader.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2020 Ben Ashbaugh
```
kpet, line 3: Should be "The Khronos Group"

scripts/gen_loader.py
```diff
@@ -0,0 +1,141 @@
...
+# noneStr - returns string argument, or "" if argument is None.
+def noneStr(s):
+    if s:
+        return s
+    return ""
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('-registry', action='store',
```
kpet, line 50: I think this option could use a more explicit name. `--api-grammar-file` (or better)?

scripts/gen_loader.py
```diff
@@ -0,0 +1,141 @@
...
+                        default='cl.xml',
+                        help='Use specified registry file instead of cl.xml')
+    parser.add_argument('-o', action='store', dest='directory',
+                        default='.',
+                        help='Create target and related files in specified directory')
+
+    args = parser.parse_args()
+
+    specpath = args.registry
+    #specpath = "https://raw.githubusercontent.com/KhronosGroup/OpenCL-Registry/master/xml/cl.xml"
```
kpet, line 60: stray comment?

kpet (commented):
This looks great! Just a few minor comments.

scripts/gen_loader.py
bashbaug: Eek, good catch.  This is the wrong licence header too.  I've fixed both problems.

scripts/gen_loader.py
bashbaug: This is the same name that's used for the spec scripts: [example](https://github.com/KhronosGroup/OpenCL-Docs/blob/master/scripts/gen_dictionaries.py#L47).

I'd prefer to stay consistent.

kpet: Not great but agree that consistency is good.

kpet (approved):

PR #115, opened 2020-08-11
Comprehensive CMake Package Config support

MathiasMagnus:
This PR is part of a set of PRs implementing CMake Package Config detection support for all OpenCL repos independently and as part of the OpenCL SDK as well.

Related PRs:
- [OpenCL-Headers](https://github.com/KhronosGroup/OpenCL-Headers/pull/107)
- [OpenCL-CLHPP](https://github.com/KhronosGroup/OpenCL-CLHPP/pull/109)

I tried keeping attribution where applicable with parallel PRs were in place.


CMakeLists.txt
```diff
@@ -130,28 +130,83 @@ if (WIN32)
...
 if (BUILD_TESTING)
     add_subdirectory (test)
 endif()
 
+install(
+  TARGETS OpenCL
+  EXPORT OpenCLICDLoaderTargets
+)
+
+include(GNUInstallDirs)
```
EwanC, line 174: We do this on line 4, is the repeated `include` necessary?

MathiasMagnus: No, removed the duplicate.

.appveyor.yml
```diff
@@ -11,10 +11,28 @@ configuration:
   - Release
 
 before_build:
-  - git clone --depth=1 https://github.com/KhronosGroup/OpenCL-Headers inc/OpenCL-Headers
-  - cmake -DOPENCL_ICD_LOADER_HEADERS_DIR=inc/OpenCL-Headers -H. -Bbuild -A%PLATFORM%
+  - git clone https://github.com/MathiasMagnus/OpenCL-Headers --branch pkgconfig --single-branch
```
alycm, line 14: It looks like this should be able to point to the `KhronosGroup` repo now?

.appveyor.yml
```diff
@@ -11,10 +11,28 @@ configuration:
...
+  - mkdir build
+  - mkdir install
+  - cd build
+  - cmake -A %PLATFORM%
+          -D CMAKE_INSTALL_PREFIX=../install
+          ..
+  - cmake --build .
+          --target INSTALL
+  - cd ..\..\
+  - git clone https://github.com/MathiasMagnus/OpenCL-ICD-Loader --branch pkgconfig --single-branch
```
alycm, line 25: Should also point to the Khronos repo, master branch.

.github/workflows/presubmit.yml
```diff
@@ -3,6 +3,7 @@ on: [push, pull_request]
 
 jobs:
   build:
+    if: false
```
alycm, line 6: I'm not familiar with GitHub actions.  Does this disable this entire build?

test/driver_stub/cl.c
```diff
@@ -171,7 +171,7 @@ clGetDeviceIDs(cl_platform_id   platform,
 {
     cl_int ret = CL_SUCCESS;
 
-    if ((num_entries > 1 || num_entries < 0) && devices != NULL) {
+    if ((num_entries > 1 /*|| num_entries < 0*/) && devices != NULL) {
```
alycm, line 174: Just delete the commented out code.

test/loader_test/icd_test_match.c
```diff
@@ -26,14 +26,16 @@ int test_icd_match()
...
+        printf("APPLOG:\n%s\n\n", app_log);
+        printf("STUBLOG:\n%s\n\n", stub_log);
         error = 1;
         goto End;
     }
 
 End:
     free(app_log);
-    free(stub_log);
+    //free(stub_log);
```
alycm, line 38: Very quickly following `test_icd_get_stub_log` it looks like this *should* be free'd?

loader/icd_dispatch_generated.c
```diff
@@ -4246,7 +4246,7 @@ CL_API_ENTRY void CL_API_CALL clSVMFree(
 {
 #if defined(CL_ENABLE_LAYERS)
     if (khrFirstLayer)
-        return khrFirstLayer->dispatch.clSVMFree(
+        khrFirstLayer->dispatch.clSVMFree(
```
alycm, line 4249: I guess the `return` is cut because `clSVMFree` returns `void`.

But now the function will continue on and call `context->dispatch->clSVMFree`.   Looking at `clCreateSamplerWithProperties` for comparison (just because it happens to be just below here), it does **not** call the `context->dispatch->clFoo` for this case.

Should there be a `return;` after this line (but adding `{}` to the if-block of course)?

test/loader_test/test_buffer_object.c
```diff
@@ -438,9 +438,10 @@ int test_buffer_object()
         test_clEnqueueCopyBufferRect(&clEnqueueCopyBufferRectData[i]);
     }
 
-    for (i=0; i<NUM_ITEMS_clEnqueueMapBuffer; i++) {
-        test_clEnqueueMapBuffer(&clEnqueueMapBufferData[i]);
-    }
+    // Segfaults inside test_icd_app_log at vfprintf
```
alycm, line 441: Well that's not encouraging.  At the least could there be an issue filed about fixing whatever is wrong here?

test/loader_test/test_image_objects.c
```diff
@@ -349,9 +349,10 @@ int test_image_objects()
         test_clEnqueueCopyBufferToImage(&clEnqueueCopyBufferToImageData[i]);
     }
 
-    for (i = 0; i<NUM_ITEMS_clEnqueueMapImage; i++) {
-        test_clEnqueueMapImage(&clEnqueueMapImageData[i]);
-    }
+    // Segfaults when freeing the returned pointer
```
alycm, line 352: Again, could there be an issue filed about fixing whatever is wrong here?  Or maybe both the segfaults are one issue.

alycm (commented):
A few comments and questions, mostly nothing to do with the core CMake changes.

loader/icd_dispatch_generated.c
Kerilk: Yes I have fixed the generators but I will suggest a fix similar to the generated files here.

loader/icd_dispatch_generated.c
```diff
@@ -4246,7 +4246,7 @@ CL_API_ENTRY void CL_API_CALL clSVMFree(
...
             context,
             svm_pointer);
```
Kerilk, line 4251:
```suggestion
    {
        khrFirstLayer->dispatch.clSVMFree(
            context,
            svm_pointer);
        return;
    }
```

Kerilk (changes requested):

loader/icd_dispatch_generated.c
Kerilk: The pull request is there: https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/139.

alycm: Ah, thank you @Kerilk, I hadn't realized that this file was generated.

.github/workflows/presubmit.yml
MathiasMagnus: Yes, but you're right, I should delete the entire file. The rework is a replacement to `presubmit.yml`.

.appveyor.yml
MathiasMagnus: I skimmed over appveyor still being there. The entire file needs to be deleted as GA has all the types of runners we covered with appveyor. (AFAIK, but do correct me if I'm wrong.)

test/loader_test/icd_test_match.c
MathiasMagnus: Yes, I missed commenting on that. I found that freeing the stub log crashed the test on my local box. Since it letting it leak and be it freed 3 statements later on `return 0;` seemed like a viable "hotfix" I went with that. The logs weren't empty, but free triggered an assertion. (MSVC, Debug build)

What should the remedy be?

loader/icd_dispatch_generated.c
MathiasMagnus: Is it okay if the generated file remains patched and the PR will overwrite it in the future?

(Ultimately we ought to generate those files on every build node; one `add_custom_target()` to the CMake scripts isn't black magic. How does one use mako to generate those headers?)

MathiasMagnus: But indeed, it was cut because initially I was set on making `-Werror` and `/WX` happy, but ultimately I gave up. One of the tested compilers was warning about returning `void`.

Kerilk: https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/139 has been merged, so maybe reverting those two (`icd_dispatch_generated.c` and `icd_print_layer_generated.c`) would avoid a conflict.

CMakeLists.txt
```diff
@@ -115,26 +109,44 @@ if (WIN32)
...
+  target_link_libraries (OpenCL PUBLIC OpenCL::Headers)
+else ()
+  find_path(OpenCL_HEADERS_DIR
+    NAMES
+      CL/cl.h OpenCL/cl.h
+    PATH_SUFFIXES
+      include
+      OpenCL/common/inc
+  )
+  target_include_directories(OpenCL SYSTEM PUBLIC ${OpenCL_HEADERS_DIR})
```
bashbaug, line 132: I wasn't able to build the ICD loader until I built and installed the OpenCLHeaders package.  Could you please check that this configuration is supported and functional?  If this is something we'd want to support - and I think it is? - do we need to add at least one CI configuration to test it?

bashbaug (changes requested):
Thanks for all of this work - this is a really great improvement!

I was able to build and successfully run the tests with a few minor changes below, mostly related to the latest headers.  Was there anything else specifically you'd like me to run regarding the ICD loader tests?

I had a lot of the same feedback as Alastair so please be sure it is addressed.

Additionally:

* I think we need to update the README since some of the steps are out-of-date or incomplete, especially regarding how to get or otherwise setup the OpenCL headers.
* These changes need a rebase.  In addition to a merge conflict they're incompatible with the latest headers.

CMakeLists.txt
```diff
@@ -110,26 +104,44 @@ if (WIN32)
...
         if (OPENCL_ICD_LOADER_PIC)
             set_target_properties(OpenCL PROPERTIES POSITION_INDEPENDENT_CODE ON)
         endif ()
-        target_link_libraries (OpenCL ${CMAKE_THREAD_LIBS_INIT})
     endif ()
 endif ()
 
-include_directories (${OPENCL_ICD_LOADER_HEADERS_DIR})
-add_definitions (-DCL_TARGET_OPENCL_VERSION=300)
+find_package(OpenCLHeaders CONFIG NO_SYSTEM_ENVIRONMENT_PATH NO_CMAKE_SYSTEM_PATH)
```
sl1pkn07, line 116: why this? this break the search when the headers(and the cmake files) is installed in the system

~~~
-- Could NOT find OpenCLHeaders (missing: OpenCLHeaders_DIR)
~~~

change to `REQUIRED` works again

sl1pkn07: seems set `-DCMAKE_FIND_USE_CMAKE_PATH=TRUE` do the trick instead of edit the cmakelist.txt

greetings and sorry for the noise

MathiasMagnus:
Superceded by #160

PR #117, opened 2020-08-16
skip testing except for icd loader project

### LINKED ISSUE #116: Disable testing when ICD Loader is included in other projects?
Currently, the ICD loader CMakefile unconditionally includes CTest, and enables tests based on BUILD_TESTING.  This is great for the ICD loader project itself, but a little awkward when the ICD loader is included in other projects, especially when the other projects also include tests.

Should we disable ICD loader testing when the ICD loader is included in other projects?

I this this is as simple as changing this:

```cmake
include (CTest)
if (BUILD_TESTING)
```

to this:

```cmake
if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
    include(CTest)
endif()
if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING)
```

More info here:

https://cliutils.gitlab.io/modern-cmake/chapters/testing.html

bashbaug:
Fixes #116 

As described in #116, this change only enables testing when building the ICD loader project itself.  If the ICD loader is included in other projects testing will not be enabled, and only the ICD loader itself will be built.


kpet (changes requested):
I think there are cases where one might want to still enable the testing even when embedding the loader (wider SDK, etc). I think we should provide a way to override this behaviour externally. `OPENCL_ICD_LOADER_ENABLE_TESTING` with defaults set as per your change?

bashbaug:
Sure, this makes sense.  I went with `OPENCL_ICD_LOADER_BUILD_TESTING` for consistency with `BUILD_TESTING`, but I'd be fine with `OPENCL_ICD_LOADER_ENABLE_TESTING` if you prefer it for some reason.

kpet (approved):
Works for me. I think this is a better name.

alycm (approved):
LGTM.

PR #119, opened 2020-08-27
Allow static build with PIC members

tpboudreau:
This patch introduces an option to allow building a static ICD Loader library with PIC members.  This is required when building a shared end-user library that will statically include the ICD Loader code (rather than a runtime dependency on a shared ICD Loader library).  An analogous build can already be achieved on WIN32 platforms by using the USE_DYNAMIC_VCXX_RUNTIME (and some other steps); this contribution enables a similar option for non-Windows builds.  Shared end-user libraries with minimal runtime dependencies are useful, for example, in building Python wheels.


bashbaug (approved):
Thank you for your contribution!

While we still discourage building the ICD loader as a static library, adding this option is reasonable and useful for use-cases that use a static library.

bashbaug:
Rebased and pushed to master as https://github.com/KhronosGroup/OpenCL-ICD-Loader/commit/b3b483303b160db080857288c1b53e8152f62b98.  Closing.

PR #122, opened 2020-09-08
Define CL_USE_DEPRECATED_OPENCL_2_{1,2}_APIS

jrprice:
We found this because some build environments were failing after #114 was merged, since it added a deprecation suffix to `clSetProgramReleaseCallback`. The actual error indicated that the GCC deprecation attribute is not valid on function definitions, so arguably there's a bug in how we're generating the loader code, but it's not a problem in practice since we just want to ignore deprecation anyway.

### DIFF
--- a/loader/icd_dispatch.h
+++ b/loader/icd_dispatch.h
@@ -35,6 +35,14 @@
 #define CL_USE_DEPRECATED_OPENCL_2_0_APIS
 #endif
 
+#ifndef CL_USE_DEPRECATED_OPENCL_2_1_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_1_APIS
+#endif
+
+#ifndef CL_USE_DEPRECATED_OPENCL_2_2_APIS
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
Makes sense.  Do we need to file an issue ~~(against the headers, perhaps)~~ to track the deprecation issue?

Edit: I misunderstood the description of the deprecation issue.  Please do file a separate issue to track the code generation issue.  Can you say a bit more what the build environment is?  Regardless, this shouldn't hold up this PR.

bashbaug:
I'm going to merge this ASAP since it looks like this is breaking CTS builds.  Separately, I'll also look at what could be going wrong with the code generation.  Thanks!

jrprice:
> Please do file a separate issue to track the code generation issue.
Will do.

> Can you say a bit more what the build environment is? Regardless, this shouldn't hold up this PR.
I was using a top-of-tree Clang with `-Werror`. It's `-Wgcc-compat` that flags this, which indicates that this might be a GCC issue too.

PR #124, opened 2020-09-09
do not add the API suffix macro to function definitions

### LINKED ISSUE #123: Function definitions shouldn't use __attribute__((deprecated))
The function definitions in the code generated for the ICD loader dispatch use the prefix/suffix preprocessor macros for marking APIs as deprecated. These macros potentially expand to include GCC's `__attribute__((deprecated))`, which isn't valid on function definitions. As noted in #122, this isn't an issue in practice for the ICD loader since we currently always support deprecated functions and therefore define the `CL_USE_DEPRECATED_OPENCL_*_APIS` macros that disable this.

bashbaug:
Fixes #123 

This PR removes the API suffix macro from all ICD dispatch functions, including generated functions and manually written functions.  The API suffix macro should be included on function headers but not on function definitions.

### DIFF
--- a/loader/icd_dispatch.c
+++ b/loader/icd_dispatch.c
@@ -150,7 +150,7 @@ static inline cl_int clGetPlatformIDs_body(
 CL_API_ENTRY cl_int CL_API_CALL clGetPlatformIDs_disp(
     cl_uint num_entries,
     cl_platform_id* platforms,
-    cl_uint* num_platforms) CL_API_SUFFIX__VERSION_1_0
+    cl_uint* num_platforms)
 {
     return clGetPlatformIDs_body(
         num_entries,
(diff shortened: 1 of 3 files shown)


alycm (approved):
LGTM, the rationale in https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/123 makes sense.

Kerilk:
We would be saving one merge if we merged/refactored this on top of the layers, as this already conflicts with `master`, and is sure to conflict with the layers.

Kerilk (approved):
LGTM

vladimirlaz:
@bashbaug  the macro has been removed from OpenCL headers (https://github.com/KhronosGroup/OpenCL-Headers/commit/9d46d9580cc469757ceb8d89842366f6c9a20093) and the build with the latest OpenCL headers failing.

Could you submit the PR?

bashbaug:
I merged https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/142 to unblock the builds, but we really should get this change merged so we don't run into similar issues again.

kpet (approved):
LGTM

bashbaug:
This has three approvals and has been open for a very long time, so merging.

PR #121, opened 2020-09-01
remove CMake option for OpenCL 3.0 provisional

bashbaug:
This is a draft PR that removes the CMake option for OpenCL 3.0 provisional and enables the OpenCL 3.0 APIs by default.

PR #125, opened 2020-09-14
Layers

Kerilk:
Here is an implementation of layers in the OpenCL Loader. A few remarks though:
 - ~~This is built on top of https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/121 and https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/126 so those will have to be merged first~~,
 - `include/cl_icd_layer.h` should be moved in the OpenCL-Headers at some point,
 - ~~`CL_LAYER_API_VERSION` (defined in `include/cl_icd_layer.h`) name is to be discussed and it's value `0x0001` is a placeholder~~,
 - You can run tests on Linux by building the loader and running: `OCL_ICD_FILENAMES=./test/driver_stub/libOpenCLDriverStub.so OCL_ICD_LAYERS=./test/layer/libPrintLayer.so ./test/loader_test/icd_loader_test`

Here is the presentation I did to present the layers to the working group:
[OpenCL Layers.pdf](https://github.com/KhronosGroup/OpenCL-ICD-Loader/files/5446471/OpenCL.Layers.pdf)
Here is a proposed extension definition for the layers:
https://github.com/Kerilk/OpenCL-Docs/blob/master/ext/cl_khr_layers.asciidoc


Kerilk:
The windows build are failing, but I don't have a windows box to fix the issue. It seems related to d3d11, but the errors are about operator overloading, so they don't make a lot of sense to me:
```
         C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(991): error C2733: 'operator ==': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
         C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(996): error C2733: 'operator !=': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
         C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1135): error C2733: 'operator ==': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
         C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1140): error C2733: 'operator !=': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
         C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1181): error C2733: 'operator ==': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
         C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1186): error C2733: 'operator !=': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
```

jenatali:
> The windows build are failing, but I don't have a windows box to fix the issue. It seems related to d3d11, but the errors are about operator overloading, so they don't make a lot of sense to me:
> 
> ```
>          C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(991): error C2733: 'operator ==': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
>          C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(996): error C2733: 'operator !=': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
>          C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1135): error C2733: 'operator ==': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
>          C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1140): error C2733: 'operator !=': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
>          C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1181): error C2733: 'operator ==': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
>          C:\Program Files (x86)\Microsoft DirectX SDK (June 2010)\Include\d3d11.h(1186): error C2733: 'operator !=': second C linkage of overloaded function not allowed [C:\projects\opencl-icd-loader\build\OpenCL.vcxproj]
> ```

Looks like this is coming from somehow including cl_icd.h from a CPP file inside an extern "C" block: https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl_icd.h#L26. The addition of the cl_icd.h includes are the only changes I can see that could be causing this.

jenatali:
> Looks like this is coming from somehow including cl_icd.h from a CPP file inside an extern "C" block: https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl_icd.h#L26. The addition of the cl_icd.h includes are the only changes I can see that could be causing this.

Which appears to be my fault:  
https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/master/loader/windows/icd_windows_apppackage.cpp#L19

I'll see if I can clean that up.

Kerilk:
I added a missing `extern "C"` block to `cl_icd_layer.h`, but that shouldn't change your diagnosis. Thanks for the quick feedback.

jenatali:
FYI @Kerilk there's a secondary problem: including `cl_icd.h` in `icd.h` causes `DXGI.h` to be included before the explicit include in `icd_windows.c`. That explicit include uses `initguid.h` to force the `IID_IDXGIFactory` constant to be defined rather than just declared. So if you want to include `cl_icd.h` in `icd.h`, you also need to rearrange those headers.

Feel free to grab the patches from #126 to fix the build.

Kerilk:
Thanks a lot for the fixes, it worked like a charm.

jenatali:
> Thanks a lot for the fixes, it worked like a charm.
> @jenatali is there some more actions I need to take regarding the secondary problem? The build passes but I wouldn't want to leave a hidden issue (that I obviously do not fully understand) behind.

No, I fixed both (the two commits in my PR). One is to stop including C++ code inside `extern "C"` and the other is rearranging headers so that `initguid.h` is first.

Kerilk:
Thanks for the confirmation.

Kerilk:
The latest change ensures that platforms are ready to be used when the layers are initialized. This allows, even though the desirability of this feature is questionable, the layer initialization function to call OpenCL functions through the dispatch table provided by the loader.

bashbaug (commented):
Thank you for getting this working, and for your patience waiting for a review!  The implementation generally looks very good and most of my questions and comments are higher-level and directional.  Regardless, this is great work and we should find a way to get this merged sooner rather than later, even if it's compiled out initially (see last comment).

* We need to start thinking about a spec for layers.  This could be an extension spec some other kind of layer documentation.  At the very least I think it would need to define:

    * The new `clGetLayerInfo` function.
    * The new `clInitLayer` function.
    * The `CL_LAYER_API_VERSION` enum.
    
    I don't think we need to define an interface for an application to know if layers are supported or enabled, but this is something we should consider also, and this might fall out naturally if we describe layers as an extension (do we need a concept of a "loader extension"?).

    I don't think a spec needs to be very formal, at least not initially, but we should have _something_ written down.

* Right now the handshaking is one-way, where the loader calls into the layer to get the layer API version.  Should there be a handshake in the other direction also, perhaps when the layer is initialized, so the layer has visibility into the loader version?

* This doesn't need to be done now, but eventually we should consider loading layers from locations other than environment variables as well.  See, for example, [Vulkan Layer Discovery](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#layer-discovery).

    * It would be especially interesting to consider new API functions to explicitly enable layers.  This would obviously need to be done via some sort of extension since it would be application-visible.

Technical questions:

* This change is inserting a check into each API function, to call into the first layer if it is present and to skip layers and do the usual call through the dispatch table otherwise.  Do you have any estimate how this might affect API performance?  I'm guessing it will be insignificant, but this would be good to verify.

* Have you considered adding the ICD loader dispatching as a "terminating layer", so there is always at least one "layer"?  In addition to (possible) code simplification this would avoid the conditional check in each API function, albeit at the cost of a couple of additional indirections:

```c
///////////////////////////////////////////////////////////////////////////////

CL_API_ENTRY cl_int CL_API_CALL clSetKernelArg(
    cl_kernel kernel,
    cl_uint arg_index,
    size_t arg_size,
    const void* arg_value) CL_API_SUFFIX__VERSION_1_0
{
    // khrFirstLayer always exists and may just be the "terminating layer":
    return khrFirstLayer->dispatch.clSetKernelArg(
        kernel,
        arg_index,
        arg_size,
        arg_value);
}

///////////////////////////////////////////////////////////////////////////////

// This is the function that is in the "terminating layer" dispatch table.
// It is otherwise exactly the same as the API function without layers:
static CL_API_ENTRY cl_int CL_API_CALL clSetKernelArg_disp(
    cl_kernel kernel,
    cl_uint arg_index,
    size_t arg_size,
    const void* arg_value) CL_API_SUFFIX__VERSION_1_0
{
    KHR_ICD_VALIDATE_HANDLE_RETURN_ERROR(kernel, CL_INVALID_KERNEL);
    return kernel->dispatch->clSetKernelArg(
        kernel,
        arg_index,
        arg_size,
        arg_value);
}
```

* Is there ever a case where a layer could be initialized multiple times (where `clInitLayer` is called multiple times), or is this invalid?  I'm not exactly sure where I'm going with this, although I have been pondering how layers would (or would not) work with something like an [OpenCL Runtime Loader](https://github.com/bashbaug/opencl-runtime-loader), which could also load and initialize layers.

* These changes are big enough that we should rev the file version (see: https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/master/loader/windows/OpenCL.rc#L23).

* How difficult would it be to add a CMake control to compile out layer support, while layers are being finalized or to eliminate the extra conditional / extra indirection?

loader/icd.c
```diff
@@ -189,6 +191,127 @@ void khrIcdVendorAdd(const char *libraryName)
...
+    const struct _cl_icd_dispatch *layerDispatch = NULL;
+    cl_uint layerDispatchNumEntries = 0;
+    cl_uint loaderDispatchNumEntries = 0;
+
+    // require that the library name be valid
+    if (!libraryName)
+    {
+        goto Done;
+    }
+    KHR_ICD_TRACE("attempting to add vendor %s...\n", libraryName);
```
bashbaug, line 213: There are a few minor copy-paste bugs in this function.

vendor --> layer

clIcdGetPlatformIDsKHR --> clInitLayer

Kerilk: Thanks Ben, nice catches.

Kerilk:
> Thank you for getting this working, and for your patience waiting for a review! The implementation generally looks very good and most of my questions and comments are higher-level and directional. Regardless, this is great work and we should find a way to get this merged sooner rather than later, even if it's compiled out initially (see last comment).

Thanks for the nice review, I will try answering your comments one by one.

>     * We need to start thinking about a spec for layers.  This could be an extension spec some other kind of layer documentation.  At the very least I think it would need to define:
>       
>       * The new `clGetLayerInfo` function.
>       * The new `clInitLayer` function.
>       * The `CL_LAYER_API_VERSION` enum.

I agree with this list.

>       I don't think we need to define an interface for an application to know if layers are supported or enabled, but this is something we should consider also, and this might fall out naturally if we describe layers as an extension (do we need a concept of a "loader extension"?).
>       I don't think a spec needs to be very formal, at least not initially, but we should have _something_ written down.

I can write a more formal documentation than the one written in the presentation, maybe it can be used as a basis for the specification.

>     * Right now the handshaking is one-way, where the loader calls into the layer to get the layer API version.  Should there be a handshake in the other direction also, perhaps when the layer is initialized, so the layer has visibility into the loader version?

You are correct that the only introspection the layer has into the loader is the size of it's dispatch table. This gives the layer the list of the supported functions of the loader but not much else. We could envision passing a structure to the clInitLayer function instead of the size of the dispatch table, and encapsulate more information inside of it.

>     * This doesn't need to be done now, but eventually we should consider loading layers from locations other than environment variables as well.  See, for example, [Vulkan Layer Discovery](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#layer-discovery).

I agree completely, I would like to have the ability to set-up system-wide layers. I will look at the Vulkan documentation.

>       * It would be especially interesting to consider new API functions to explicitly enable layers.  This would obviously need to be done via some sort of extension since it would be application-visible.

There are some interesting synchronization issues here that need to be well understood before we can proceed, and it may need additional APIs inside the layers themselves. That's why when discussing it with Kevin we decided it was maybe not a good idea for a first shot.

> Technical questions:
> 
>     * This change is inserting a check into each API function, to call into the first layer if it is present and to skip layers and do the usual call through the dispatch table otherwise.  Do you have any estimate how this might affect API performance?  I'm guessing it will be insignificant, but this would be good to verify.

We could try to measure the overhead of a single if, but I counted on the branch predictor to remove the overhead rapidly, at least on the frequently used APIs

>     * Have you considered adding the ICD loader dispatching as a "terminating layer", so there is always at least one "layer"?  In addition to (possible) code simplification this would avoid the conditional check in each API function, albeit at the cost of a couple of additional indirections:

In the proposal, the ICD loader dispatching is the "terminating layer". I only deactivate it if there is no layer loaded.
I did this to favor performance, given that most of the code is generated, the additional complexity is, I think, manageable.



>     * Is there ever a case where a layer could be initialized multiple times (where `clInitLayer` is called multiple times), or is this invalid?  I'm not exactly sure where I'm going with this, although I have been pondering how layers would (or would not) work with something like an [OpenCL Runtime Loader](https://github.com/bashbaug/opencl-runtime-loader), which could also load and initialize layers.

This is discussed a bit in the presentation and in the code comments. For now, the loader will refuse to load a layer a second time, if it correspond to the same inode in the file system (I don't know enough of Windows libraries to know how they would behave in this case, but from what I read it behaves similarly). If the same library is copied to a different file, the the dlopen function will return a different handle and instanciate the library in the program memory space but at different addresses, and they won't share their data, in this way you can use several time the same layer.

You can also force the same library to be loaded in a new memory space using the LM_ID_NEWLM flag with dlopen, but this would complexify the code, as these memory space are limited and you would need some bookkeeping to know in which memory space the library was previously instanciated.
 
>     * These changes are big enough that we should rev the file version (see: https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/master/loader/windows/OpenCL.rc#L23).

Indeed. Should this revision occur in this pull request or subsequently?

>     * How difficult would it be to add a CMake control to compile out layer support, while layers are being finalized or to eliminate the extra conditional / extra indirection?

I will add a macro to guard the relevant sections, so the layers can be deactivated if need be.

Kerilk:
I added a first draft of an extension definition.
https://github.com/Kerilk/OpenCL-Docs/blob/master/ext/cl_khr_layers.asciidoc
@bashbaug please tell me if it corresponds to what you had in mind.

bashbaug (commented):
Mostly LGTM - thanks!

I only have two small questions/comments:

* I still think we should bump the file version for Windows from `3.0.0` to `3.0.1` in OpenCL.rc.

* Should the layer support be enabled by default while the feature is being formalized (this is the current behavior), or should be opt-in for now?

Let's discuss both of these on the call today.

Kerilk:
Thanks Ben for the feedback.

> Let's discuss both of these on the call today.

Unfortunately we didn't get there.
 - These two issues are linked, what is the point of bumping the file version if the feature is not active in it? That's why I was proposing the version increment be done in a separate pull request.
 - I am worried about the layer not being tested by the continuous integration system if they are disabled by default.

I would also like at least another set of eyes on the pull request, I am worried we may have missed something, given how big the changes are.

Kerilk:
I added a section describing the state of the layer support in the README file.

Kerilk:
The last blocking issue (enum value) is resolved. This should be mergeable now.

Kerilk:
I updated the Windows version number.

Kerilk:
Fixed the MacOS build too.

loader/icd.c
```diff
@@ -189,6 +195,129 @@ void khrIcdVendorAdd(const char *libraryName)
...
+        goto Done;
+    }
+
+    if (CL_LAYER_API_VERSION_100 != api_version)
+    {
+        KHR_ICD_TRACE("unsupported api version\n");
+        goto Done;
+    }
+
+    layer = (struct KHRLayer*)calloc(sizeof(struct KHRLayer), 1);
```
alycm, line 267: Valgrind isn't complaining about leak, but I can't figure out how this gets free'd for the "success" path.  The "failure" path is clearly handled in `Done`.  I think valgrind is happy because it's still reachable, and it is never free'd because it is required until the program is done?

alycm (approved):
To try this out I wrote a tiny layer to "validate" something as that seems a common use case for layers.  It worked and it wasn't difficult to write or use!

Just some thoughts from trying this out.  I do NOT consider these blockers to merging, they all come down to either (a) documenting how to use layers, or (b) specifying what a layer is implementing.  For an experimental feature both can clearly come later.

* If `OCL_ICD_LAYERS` contains an invalid path this will be silently ignored.
* If the .so pointed to does not contain `clGetLayerInfo` and `clInitLayer` symbols then the layer will be silently ignored.
* If `clGetLayerInfo` can't handle `CL_LAYER_API_VERSION` then the layer will be silently ignored.
* Changing the `#if 0` around `KHR_ICD_TRACE` macro definition makes these conditions non-silent, useful when writing a layer!
* `CL_LAYER_API_VERSION` should return a type of `cl_layer_api_version`.
* Are `clInitLayer` and `clGetLayerInfo` meant to be threadsafe?  Whether this matters depends on the layer of course, but the example `clInitLayer` updates dispatch structs for example.

loader/icd.c
Kerilk: Yes we don't have a defined way to unload and cleanup everything when the laoder ffinishes, so for now those questions are left unanswered for the layers as well.

include/cl_icd_layer.h
```diff
@@ -0,0 +1,62 @@
...
+#include <CL/cl.h>
+#include <CL/cl_icd.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef cl_uint cl_layer_info;
+typedef cl_uint cl_layer_api_version;
+#define CL_LAYER_API_VERSION 0x4240
```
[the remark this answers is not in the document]
Kerilk: My take exactly, I expect this header to either be merged inside the global one, or be migrated as is, like the cl_icd.h is.

include/cl_icd_layer.h
```diff
@@ -0,0 +1,62 @@
...
+#define CL_LAYER_API_VERSION_100 100
```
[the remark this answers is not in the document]
Kerilk: I will make sure this is better explained in:
https://github.com/Kerilk/OpenCL-Docs/blob/master/ext/cl_khr_layers.asciidoc
I should submit this pull request as soon as this one is merged.

Kerilk:
Thanks @alycm for the review!

I answered some of your comments inline, here is the rest:

> - If OCL_ICD_LAYERS contains an invalid path this will be silently ignored.
> - If the .so pointed to does not contain clGetLayerInfo and clInitLayer symbols then the layer will be silently ignored.
> - If clGetLayerInfo can't handle CL_LAYER_API_VERSION then the layer will be silently ignored.
> - Changing the #if 0 around KHR_ICD_TRACE macro definition makes these conditions non-silent, useful when writing a layer!

Yes this is the same design choice that has been made while loading drivers (the code is "very" similar (some would say copy pasted :-)).

> Are clInitLayer and clGetLayerInfo meant to be threadsafe? Whether this matters depends on the layer of course, but the example clInitLayer updates dispatch structs for example.

The fact is that they are only called from inside the critical section of the loader initialization. From my point of view it means only one thread should execute all of those, and the OpenCL calls some of them may wish to do during those calls. I haven't found a way to fault this reasoning but I may just be lacking imagination, I guess at this point time will tell if we manage to get race conditions there.

Thanks again.

alycm:
@Kerilk Sorry, despite the link being in the PR description I completely missed that https://github.com/Kerilk/OpenCL-Docs/blob/master/ext/cl_khr_layers.asciidoc is already written.  That would have answered some of my questions I think.

bashbaug (approved):
LGTM!  Merging as we discussed in the December 8th teleconference.

PR #132, opened 2020-12-04
Switch from Travis CI to Github Actions

jrprice:
It's pretty basic, but offers the same level of coverage as the existing Travis CI config, with the benefit that it actually runs. I'll open some issues for adding additional coverage (e.g. Armv7, AArch64, Clang, Windows).

I will enable macOS in #131 when I rebase it on top of this.


jrprice:
Merging immediately to get CI moving again.

PR #135, opened 2020-12-10
Add C99 flag as last commit added some c99 variable declarations

JohnKesapidesARM:
#131 added some c99 variables declarations. 
This commit adds the C99 req in CMAKE
Signed-off-by: John Kesapides <john.kesapides@arm.com>


kpet (changes requested):
You probably want to set `CMAKE_C_STANDARD_REQUIRED` as well.

kpet (approved):
LGTM

PR #139, opened 2021-02-04
Fixed generators to better account for the lack of return value of clSVMFree.

Kerilk:
As raised by https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/115.

### DIFF
--- a/scripts/icd_dispatch_generated.c.mako
+++ b/scripts/icd_dispatch_generated.c.mako
@@ -78,7 +78,12 @@ ${("", "static ")[disp]}CL_API_ENTRY ${api.RetType} CL_API_CALL ${api.Name + (""
 %if disp == 0:
 #if defined(CL_ENABLE_LAYERS)
     if (khrFirstLayer)
+%  if api.Name == "clSVMFree":
+    {
+        khrFirstLayer->dispatch.${api.Name}(
+%  else:
         return khrFirstLayer->dispatch.${api.Name}(
+%  endif
 %for i, param in enumerate(api.Params):
 %  if i < len(api.Params)-1:
             ${param.Name},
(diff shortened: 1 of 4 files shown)


scripts/icd_dispatch_generated.c.mako
```diff
@@ -78,14 +78,23 @@ ${("", "static ")[disp]}CL_API_ENTRY ${api.RetType} CL_API_CALL ${api.Name + (""
 %if disp == 0:
 #if defined(CL_ENABLE_LAYERS)
     if (khrFirstLayer)
+%  if api.Name == "clSVMFree":
```
bashbaug, line 81: This is fine for now, but if we wanted to make this a little more general we could check for a `void` return type here vs. `clSVMFree` specifically.

bashbaug (commented):
Thanks for fixing this!

I think we'll need to generate new files now also?  Otherwise LGTM.

Kerilk:
The fixed generated files are already in #115. I would like to avoid a conflict. What do you think?

scripts/icd_dispatch_generated.c.mako
Kerilk: Yes it occurred to me, but I would have had to make this change several times throughout the file, so I kept it uniform. We will need to refactor it at some point.

bashbaug:
I think we should try to keep the generated file in sync with the template, in which case the generated file should be included in this PR also.  I'll work on adding some automation to enforce this.

Kerilk:
I updated the generated files. Thanks for the feedback Ben.

bashbaug (approved):
Thanks - LGTM.  Merging ASAP to unblock #115.

PR #142, opened 2021-02-21
Replace uses of CL_EXT_{PRE,SUF}FIX* with CL_API_{PRE,SUF}FIX*

kpet:
Generation done using https://github.com/KhronosGroup/OpenCL-Docs/pull/569

Signed-off-by: Kévin Petit <kpet@free.fr>


bashbaug (approved):
LGTM.

We should really merge #124 to remove most (all?) of these suffix defines but merging this first to get the builds working again.

PR #144, opened 2021-03-12
Remove cl_gl_ext.h includes

jrprice:
All OpenGL extensions are now in cl_gl.h.


bashbaug (approved):
LGTM.

There is more opportunity for include file cleanup in the future.  As one example, we could collapse many (most? almost all?) of these include by using `opencl.h` instead, and I don't think the ICD loader needs to know anything about the EGL and DX interop extensions.  Both of these can be done by a separate PR.

PR #145, opened 2021-03-26
Change OCL_ICD_LAYERS to OPENCL_LAYERS to conform to the specification.

Kerilk:
This looks like an oversight on my part:

https://github.com/KhronosGroup/OpenCL-Docs/blob/master/ext/cl_loader_layers.asciidoc

states:

> 
>  ## New Environment Variables
> 
> A colon separated list of layer libraries that the Loader must try to load. The layers are loaded in the specified order. The most recently loaded layer will be first in the dispatch chain.
> 
> `OPENCL_LAYERS`

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ configuration by using the `ENABLE_OPENCL_LAYERS` (`ON` by default) cmake variab
 cmake -DENABLE_OPENCL_LAYERS=OFF
 ```
 
-For now, runtime configuration of layers is done using the `OCL_ICD_LAYERS` environment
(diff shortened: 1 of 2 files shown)


Kerilk:
It was switched during the discussions because it was decided ICD was confusing in this context. If we feel it is too late to change it we need to modify the specification.

bashbaug:
I don't think it's too late to change the name, but if we're going to change it we should do so ASAP.

kpet (approved):
LGTM!

bashbaug (approved):
Merging as discussed in the March 30th teleconference.

Issue #146, opened 2021-04-15
Broken for Arch Linux


andrfgs:
ICD Loader seems to be broken on Arch Linux due to multiple linker definitions. Here's the log compiling v2020.06.16:

```
/usr/bin/ld: CMakeFiles/icd_loader_test.dir/test_platforms.c.o:(.bss+0x110): multiple definition of `ret_val'; CMakeFiles/icd_loader_test.dir/test_kernel.c.o:(.bss+0x348): first defined here
/usr/bin/ld: CMakeFiles/icd_loader_test.dir/test_program_objects.c.o:(.bss+0x0): multiple definition of `ret_val'; CMakeFiles/icd_loader_test.dir/test_kernel.c.o:(.bss+0x348): first defined here
/usr/bin/ld: CMakeFiles/icd_loader_test.dir/test_sampler_objects.c.o:(.bss+0x0): multiple definition of `ret_val'; CMakeFiles/icd_loader_test.dir/test_kernel.c.o:(.bss+0x348): first defined here
/usr/bin/ld: CMakeFiles/icd_loader_test.dir/test_cl_runtime.c.o:(.bss+0x0): multiple definition of `ret_val'; CMakeFiles/icd_loader_test.dir/test_kernel.c.o:(.bss+0x348): first defined here
/usr/bin/ld: CMakeFiles/icd_loader_test.dir/test_clgl.c.o:(.bss+0x248): multiple definition of `ret_val'; CMakeFiles/icd_loader_test.dir/test_kernel.c.o:(.bss+0x348): first defined here
/usr/bin/ld: CMakeFiles/icd_loader_test.dir/test_image_objects.c.o:(.bss+0x0): multiple definition of `ret_val'; CMakeFiles/icd_loader_test.dir/test_kernel.c.o:(.bss+0x348): first defined here
collect2: error: ld returned 1 exit status
make[3]: *** [test/loader_test/CMakeFiles/icd_loader_test.dir/build.make:275: ../bin/icd_loader_test] Erro 1
make[3]: a sair da pasta "/home/user/.cache/yay/khronos-ocl-icd/src/icd/build"
make[2]: *** [CMakeFiles/Makefile2:230: test/loader_test/CMakeFiles/icd_loader_test.dir/all] Erro 2
make[2]: a sair da pasta "/home/user/.cache/yay/khronos-ocl-icd/src/icd/build"
make[1]: *** [Makefile:101: all] Erro 2
[...]

PR #101, opened 2020-03-27
Make ret_val and ret_mem local to the functions in which they are used.

ormshaw:
Make ret_val and ret_mem local to the functions in which they are used.
Also consistently used cl_int and cl_mem

PR #147, opened 2021-04-17
declare ret_val global variable static

bashbaug:
This is a possible fix for #146.

It would be better to remove all global ret_val variables, but this would be a much bigger change.

### DIFF
--- a/test/loader_test/test_kernel.c
+++ b/test/loader_test/test_kernel.c
@@ -15,7 +15,7 @@ extern cl_event event;
 extern cl_context  context;
 extern cl_command_queue command_queue;
 extern cl_device_id devices;
-int ret_val;
+static int ret_val;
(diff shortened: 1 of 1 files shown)


alycm (approved):
As this is the only surviving global `ret_val` I'm not sure this affects #146, but LGTM regardless as it makes the intent clearer.

I'm not sure why this variable was skipped in this PR: https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/101

Though I suppose once there is only one left the linker errors go away.  I'm fine with this one liner rather than duplicating the variable into each function.

bashbaug:
I think what I'd like to do is:

* Merge this as-is.  It doesn't look right to me even if it isn't causing problems in some configurations.
* Make the bigger change and duplicate the variable into each function... unless we decide to go in a different route for ICD loader unit testing instead (#76).

alycm:
Sounds good to me.

PR #148, opened 2021-05-04
Fix compiler warnings

johnplate:
This fixes compiler warnings for a custom build configuration.

### DIFF
--- a/loader/windows/icd_windows.c
+++ b/loader/windows/icd_windows.c
@@ -151,7 +151,7 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
             DWORD dwValueSize = sizeof(dwValue);
 
             // read the value name
-            KHR_ICD_TRACE("Reading value %lu...\n", dwIndex);
+            KHR_ICD_TRACE("Reading value %"PRIuDW"...\n", dwIndex);
             result = RegEnumValueA(
                   platformsKey,
                   dwIndex,
@@ -164,7 +164,7 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
             // if RegEnumKeyEx fails, we are done with the enumeration
             if (ERROR_SUCCESS != result)
             {
-                KHR_ICD_TRACE("Failed to read value %lu, done reading key.\n", dwIndex);
+                KHR_ICD_TRACE("Failed to read value %"PRIuDW", done reading key.\n", dwIndex);
                 break;
             }
             KHR_ICD_TRACE("Value %s found...\n", cszLibraryName);
--- a/loader/windows/icd_windows.c
+++ b/loader/windows/icd_windows.c
@@ -151,7 +151,7 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
             DWORD dwValueSize = sizeof(dwValue);
 
             // read the value name
-            KHR_ICD_TRACE("Reading value %d...\n", dwIndex);
+            KHR_ICD_TRACE("Reading value %lu...\n", dwIndex);
             result = RegEnumValueA(
                   platformsKey,
                   dwIndex,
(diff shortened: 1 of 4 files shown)


Kerilk:
Thanks for discovering this. I am unsure this fixes the issue fully though. Maybe we would be better off using something like:
https://stackoverflow.com/a/16949648

johnplate:
> Thanks for discovering this. I am unsure this fixes the issue fully though. Maybe we would be better off using something like:
> https://stackoverflow.com/a/16949648

You are right. Not important, we can close the PR. The warnings only showed because I forgot to disable them as we usually do for third party libs. Maybe someone will feel inspired to fully fix it.

Kerilk:
I will give it a shot, starting from your fork if that's OK with you.

johnplate:
Sure, thanks, all yours :)

Kerilk:
For those interested in following the discussion, I made a tentative proposal here: https://github.com/johnplate/OpenCL-ICD-Loader/pull/1

johnplate:
All printf format strings are fine as they are, because there are no explicit 64 bit types involved. The PR is ready for review.

Kerilk:
You are correct in the case of MSVC on windows (an `LLP64` compiler). But I think we have users that use gcc (an `LP64` compiler) to compile the Windows version of the loader. The `l` length specifier should be used on 32 bit architectures, on `LLP64` compilers like MSVC, but not on `LP64` compilers. The macros the macros proposed in:
https://github.com/Kerilk/OpenCL-ICD-Loader/commit/663a35a0409a76e6a903aacf1ee6de5db96c13b3
would enforce this behavior.

johnplate:
The conversion specifier "u" with length modifier "l" expects an argument type of `unsigned long`, see [printf spec](https://en.cppreference.com/w/c/io/fprintf).  That's C99 standard, it doesn't matter which architecture, compiler or platform. Please try it if you don't believe me. :)

Kerilk:
I did, I installed cygwin on my windows desktop and compiled this program:
```c
#include <windows.h>
#include <stdio.h>
int main() {
  DWORD d;
  unsigned long ul;
  unsigned long long ull;
#ifdef _LP64
  printf("LP64\n");
#endif
#ifdef _LLP64
  printf("LLP64\n");
#endif
  printf("DWORD: %zu\nunsigned long: %zu\nunsigned long long: %zu\n", sizeof(DWORD), sizeof(ul), sizeof(ull));
  return 0;
}
```
Compiled it with `gcc`, ran it and obtained the following output:
```
LP64
DWORD: 4
unsigned long: 8
unsigned long long: 8
```
So on this platform, `windows.h` does not define `DWORD` as an `unsigned long`. The fix I propose would work in this case also.
I hope this clears things up.

johnplate:
This is a completely different problem then what you talked about until now. I said "lu" is the correct format string for `unsigned long`, and that's true. I also mentioned that `DWORD` is always defined as `unsigned long`, and this is where I was wrong, because I didn't think of cygwin/mingw Windows headers. Sorry for that. Assumptions are always dangerous. :)

Now you assume that `DWORD` is always defined as `unsigned int` if `_LP64` is defined, and as `unsigned long` otherwise? Where is that documented?

Kerilk:
Window documentation define `DWORD` and `ULONG` range of values very specifically:
https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types

DWORD:
> A 32-bit unsigned integer. The range is 0 through 4294967295 decimal.

LONG:
> A 32-bit signed integer. The range is -2147483648 through 2147483647 decimal.

ULONG:
> An unsigned LONG. The range is 0 through 4294967295 decimal.

johnplate:
Yes, `DWORD` is defined as 32 bit. If `_LP64` is defined, only `unsigned int` is 32 bit, so `DWORD` must be `unsigned int`. How do we know that `DWORD` is `unsigned long` in all other cases?

johnplate:
We have no guarantee that `DWORD` is defined as `unsigned long` if `_LP64` is not defined, since `DWORD` could be also `unsigned int`, which is typically also 32 bit. I had trouble before compiling Khronos headers on Windows with clang for arm64. There might always be more development environments than someone is aware of.

For now your solution is the best that I could find, so we should use it. If a non-LP64 system shows up which doesn't define `DWORD` as `unsigned long`, the macros have to be adjust for that system. The only alternatives for now would be to always cast `DWORD` to `unsigned long` and simply use format string "lu", or to evaluate the type of `DWORD` in a configuration step and adjust the format string macros accordingly.

Kerilk:
Thanks for bringing this issue to our attention, I'll let other Khronos members review the proposed changes, maybe they will have more insights on the problem.

bashbaug (approved):
LGTM.  The changes are almost all within the KHR_ICD_TRACE macro, which compiles to nothing by default, so the risk to most code is small (I verified a build with KHR_ICD_TRACE enabled just in case).

It would be nice to use the `<inttypes.h>` macros instead of "rolling our own" for this.  Was this investigated?  The StackOverflow link hints that some compilers may not be supporting `<inttypes.h>` yet, but the comment was from 2013 so hopefully they've caught up by now? 
 This can be addressed in a subsequent PR, just curious.

I'll leave this open for another day just in case there are any other comments, but I'll merge first thing tomorrow if there are no additional comments.  Thanks!

PR #149, opened 2021-05-26
Add support for QNX build environment and make Linux-only functions more POSIX-like.

mgorchak-blackberry:
Hello! 

This PR consist of two patches, first adds "build-qnx" folder to the project to provide support for QNX platform, second brings more POSIX like behavior to the linux platform functions what makes them usable under any POSIX compliant OS. Not all operation systems have extended "dirent" structure to obtain filesystem entries details and type.

Thank you!

### DIFF
--- a/loader/linux/icd_linux.c
+++ b/loader/linux/icd_linux.c
@@ -27,6 +27,16 @@
 #include <dirent.h>
 #include <pthread.h>
 
+#if !defined(DT_REG)
+/* Try universal POSIX way if DT_REG definition is not available */
+#if !defined(_POSIX_C_SOURCE)
+#define _POSIX_C_SOURCE 1
+#endif /* _POSIX_C_SOURCE */
+/* For _POSIX_PATH_MAX */
+#include <limits.h>
+#include <sys/stat.h>
+#endif /* !DT_REG */
+
 static pthread_once_t initialized = PTHREAD_ONCE_INIT;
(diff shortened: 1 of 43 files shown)


bashbaug:
Hi @mgorchak-blackberry, thank you for your contribution (and for your patience)!  We discussed this PR in our weekly teleconference and our biggest concern is maintaining this build environment.  Would it be possible to add this configuration to our CI?  If not, we can merge this PR and we'll make our best effort not to break it, but it's possible we'll inadvertantly introduce a build break because it's not something we'll test regularly.  Thanks!

loader/linux/icd_linux.c
```diff
@@ -61,11 +71,34 @@ void khrIcdOsVendorsEnumerate(void)
         // attempt to load all files in the directory
         for (dirEntry = readdir(dir); dirEntry; dirEntry = readdir(dir) )
         {
+#if defined(DT_REG)
```
bashbaug, line 74: This change is a bit hard to follow because it breaks up a switch statement with an `#ifdef`.  Would it be possible to refactor this code slightly to increase readability / maintainability?  I think this could look something like: Initialize a flag to "false", set it to "true" in the `#ifdef` code when appropriate, then check the flag outside of the `#ifdefs`.

bashbaug (changes requested):

PR #150, opened 2021-06-12
update ICD loader generation scripts to handle empty arguments

bashbaug:
See: https://github.com/KhronosGroup/OpenCL-Docs/pull/593

These are the most minimal changes needed to handle APIs with no arguments in the XML file.  I tried various other solutions that preserved empty arguments and handled this case during generation, but it turned out to be much easier to just add a dummy "void" argument instead and leave the rest of the scripts unchanged.

We can investigate a different solution if needed but this should unblock the OpenCL-Docs PR mentioned above.

Tagging @Kerilk for review.

### DIFF
--- a/scripts/gen/__init__.py
+++ b/scripts/gen/__init__.py
@@ -92,6 +92,12 @@ def get_apisigs(spec):
             ptypeend = ptypeend.strip()
             pname = pname.strip()
             plist.append(ApiParam(ptype, ptypeend, pname))
+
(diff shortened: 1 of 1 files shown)


Kerilk:
Thanks for this. This is the simplest and best way to go about it for now.

Kerilk (approved):

bashbaug:
Since this is purely an update to the code generation toolchain and has a +1 I'm going to merge it to unblock https://github.com/KhronosGroup/OpenCL-Docs/pull/593.

PR #152, opened 2021-07-27
Supported an OPENCL_VISIBLE_DEVICES environment variable inspired fro…

cdai2:
…m CUDA_VISIBLE_DEVICES.

OPENCL_VISIBLE_DEVICES=descriptor_list
descriptor_list -> descriptor | descriptor::descriptor_list
descriptor -> driver,platform | driver,platform,type | driver,platform,type,device_list
device_list -> device_id | device_id,device_list
driver: client driver (foo.icd or foo.so)
platform: platform ID
type: gpu, cpu, accelerator, custom, or any
device_id: device ID in the platform

This patch is to land https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/45 again.
Most of the code is from pull request 45.

e.g., OPENCL_VISIBLE_DEVICES=foo.icd,0,gpu,0,1::bar.icd,1,cpu,0 only shows
- GPU 0, 1 in the first platform of foo.icd
- CPU 0 in the second platform of bar.icd

### DIFF
--- a/loader/icd_dispatch_generated.c
+++ b/loader/icd_dispatch_generated.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012-2020 The Khronos Group Inc.
+ * Copyright (c) 2012-2021 The Khronos Group Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,9 @@
  */
 
 #include "icd_dispatch.h"
+#include "icd_envvars.h"
 #include "icd.h"
+#include "icd_device_visible.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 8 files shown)


loader/icd_dispatch_generated.c
```diff
@@ -96,12 +98,81 @@ CL_API_ENTRY cl_int CL_API_CALL clGetDeviceIDs(
             num_devices);
 #endif // defined(CL_ENABLE_LAYERS)
     KHR_ICD_VALIDATE_HANDLE_RETURN_ERROR(platform, CL_INVALID_PLATFORM);
-    return platform->dispatch->clGetDeviceIDs(
```
jenatali, line 99: This is a generated file. You don't want to modify this, you want to modify the generator.

jenatali (commented):
It seems to me that this would be more appropriate as a layer, now that those are available, rather than building it straight into the loader, but perhaps someone with more expertise on layers would like to comment.

loader/icd_dispatch_generated.c
cdai2: sorry.  I closed this PR by mistake and I created PR 153. Please review.
I updated the icd_dispatch_generated.c file.

PR #154, opened 2021-08-11
System layer implementation.

Kerilk:
This is a tentative implementation of system wide layers. Those layers are supposed to always be loaded before user-specified layers, meaning they will be deeper in the layer stack.

On Linux/MacOs/Android:
 - By default layers are configured in `/etc/OpenCL/layers/` (`/system/vendor/Khronos/OpenCL/layers` on Android);
 - This path can be overiden using the `OPENCL_LAYER_PATH` environment variable (similar to `OCL_ICD_VENDORS`);
 - Similarly to ICDs, layers are configured in a text file ending in `.lay`;
 - As ordering of layers is important, they are alphanumerically sorted by filename, meaning that prefixing `.lay` files with priorities should allow the desired ordering (akin to many configuration systems).

On Windows:
 - Layers are configured through the registry key: `SOFTWARE\Khronos\OpenCL\Layers`;
 - Similarly to ICDs, the layers are added through a `REG_DWORD` values;
 - Contrary to ICDs, the value of the keys can be non zero and will be used as a priority value to sort them.

Many of these choices are arbitrary and can be discussed, so feel free to rip through this proposal.

I tried to respect each system coding practices but may have missed the mark. Especially, for the windows code path, I didn't feel confident enough to refactor the two variable list implementations and come up with something satisfactory. I took more liberties refactoring the *nix code path.

I extensively tested the Linux implementation by generating different print layers with slightly modified output and trying different ordering and combination of system and user specified layers (using the `OPENCL_LAYERS` environment variable), but the diff is quite big so I may have let slip something through. I could only validate the windows code path through building it using GitHub actions, so it would be good that it is given a thorough look/testing.

PS: I am unsure I am correctly understanding the Windows safe mode issue: as far as I understand it prohibits the usage of environment variables, preventing any user configuration of the loader through this mean. Is that correct? If it is it would prevent any device filtering using environment variables unless some variables are treated differently.


loader/linux/icd_linux.c
```diff
@@ -141,7 +200,19 @@ void khrIcdOsVendorsEnumerate(void)
...
+
+// go through the list of vendors in the two configuration files
+void khrIcdOsVendorsEnumerate(void)
+{
+    khrIcdVendorsEnumerateEnv();
+
+    khrIcdOsDirEnumerate(ICD_VENDOR_PATH, "OCL_ICD_VENDORS", ".icd", khrIcdVendorAdd, 0);
+
 #if defined(CL_ENABLE_LAYERS)
+    // system layers should be closer to the driver
```
jlewis-austin, line 213: Is it important to to distinguish these as system layers? I'm thinking that a user can choose to use the file-based mechanism for their own layers if they want to, e.g. if using environment variables is inconvenient on Android.

README.md
```diff
@@ -142,4 +142,5 @@ The following debug environment variables are available for use with the OpenCL
 |:---------------------------------:|---------------------|----------------------|
 | OCL_ICD_FILENAMES                 | Specifies a list of additional ICDs to load.  The ICDs will be enumerated first, before any ICDs discovered via default mechanisms. | `export OCL_ICD_FILENAMES=libVendorA.so:libVendorB.so`<br/><br/>`set OCL_ICD_FILENAMES=vendor_a.dll;vendor_b.dll` |
 | OCL_ICD_VENDORS                   | On Linux and Android, specifies a directory to scan for ICDs to enumerate in place of the default `/etc/OpenCL/vendors'. |  `export OCL_ICD_VENDORS=/my/local/icd/search/path` |
-| OPENCL_LAYERS                    | Specifies a list of layers to load. |  `export OPENCL_LAYERS=libLayerA.so:libLayerB.so`<br/><br/>`set OPENCL_LAYERS=libLayerA.dll;libLayerB.dll` |
+| OPENCL_LAYERS                     | Specifies a list of layers to load. |  `export OPENCL_LAYERS=libLayerA.so:libLayerB.so`<br/><br/>`set OPENCL_LAYERS=libLayerA.dll;libLayerB.dll` |
+| OCL_LAYERS                        | On Linux and Android, specifies a directory to scan for layers to enumerate in place of the default `/etc/OpenCL/layers'. | `export OCL_ICD_VENDORS=/my/local/layers/search/path` |
```
jlewis-austin, line 146: OCL_LAYERS vs OPENCL_LAYERS are going to be confusing to remember. Maybe OCL_LAYER_LIBS and OCL_LAYER_DIR (assuming it's not too late to change the existing name)?

Kerilk:
@jlewis-austin Thanks for the feedback, it is greatly appreciated.

loader/linux/icd_linux.c
Kerilk: Indeed the user can use them, as long as they can install the libraries in the default folder. On a classic linux distribution that would require root access, unless the user overrides the default location through environment variables. This naming was sparked from a discussion about layers that could be configured in safe mode as well on windows platform, so with a system wide scope and independent from users. We can discuss giving them a new name.

README.md
Kerilk: Good catch on the copy paste, I'll fix that.
For the naming of the environment variables, you are correct that it is going to be difficult to converge on something satisfactory. The original name of OPENCL_LAYERS was OCL_ICD_LAYERS, but was changed by request of the working group. Unfortunately, rolling it back/changing it is not possible anymore. We can still change OCL_LAYERS to OCL_LAYER_LIBS or OPENCL_LAYER_LIBS.

Kerilk:
Non trivial rebase on top of master. Should be tested thoroughly.

loader/linux/icd_linux.c
```diff
@@ -36,111 +37,168 @@ static pthread_once_t initialized = PTHREAD_ONCE_INIT;
...
 {
-    DIR *dir = NULL;
-    struct dirent *dirEntry = NULL;
-    const char* vendorPath = ICD_VENDOR_PATH;
-    char* envPath = NULL;
+    struct stat statBuff;
+    char* fileName = NULL;
 
-    khrIcdVendorsEnumerateEnv();
+    // make sure the file name ends in .icd
```
MathiasMagnus, line 48: This comment became out-of-sync.
```suggestion
    // make sure the file name ends in `extension` (eg. .icd, or .lay)
```

loader/windows/icd_windows.c
```diff
@@ -97,6 +98,77 @@ void adapterFree(WinAdapter *pWinAdapter)
...
+            newCapacity = 1;
+        }
+        else if(newCapacity < UINT_MAX/2)
+        {
+            newCapacity *= 2;
+        }
+
+        WinLayer* pNewBegin = malloc(newCapacity * sizeof(*pWinLayerBegin));
+        if (!pNewBegin)
+            result = FALSE;
```
MathiasMagnus, line 136: If we print out-of-memory errors on Linux (in which case printing to the console is problematic on it's own), we should strive to be consistent and do the same on Windows.
```suggestion
        if (!pNewBegin)
        {
            KHR_ICD_TRACE("Failed allocate space for Layers array\n");
            result = FALSE;
        }
```

loader/windows/icd_windows.c
```diff
@@ -97,6 +98,77 @@ void adapterFree(WinAdapter *pWinAdapter)
...
+            pWinLayerEnd = pNewBegin + oldCapacity;
+            pWinLayerBegin = pNewBegin;
+        }
+    }
+    if (pWinLayerEnd != pWinLayerCapacity)
+    {
+        size_t nameLen = (strlen(szName) + 1)*sizeof(szName[0]);
+        pWinLayerEnd->szName = malloc(nameLen);
+        if (!pWinLayerEnd->szName)
+            result = FALSE;
```
MathiasMagnus, line 154:
```suggestion
        if (!pWinLayerEnd->szName)
        {
            KHR_ICD_TRACE("Failed allocate space for Layer file path\n");
            result = FALSE;
        }
```

MathiasMagnus (changes requested):
Overall LGTM, minor changes noted, but overall it just needs to be rebased on latest master (which since received the SDK base changes, but should apply cleanly, GitHub may even do it on it's own).

Kerilk:
> Overall LGTM, minor changes noted, but overall it just needs to be rebased on latest master (which since received the SDK base changes, but should apply cleanly, GitHub may even do it on it's own).

Thanks for the review. I've rebased the branch on `master`, so if you want to use it in the Layers CI you can do so, and it should be easy to switch back to the `master` branch once those changes have been merged. Especially, testing the layer sorting algorithm for windows could be useful. I did test the lexicographic sorting in Linux though, but Window's version works using the value of the layers in the registry.

Kerilk:
Re-based on main.

MathiasMagnus (approved):

README.md
Kerilk: I implemented the change @bashbaug suggested, as it seems a good choice.

loader/linux/icd_linux.c
bashbaug: Can we check what Vulkan is doing to see if we are consistent?

Kerilk:
I know I said otherwise at the working group meeting, but thoroughly testing the feature (variables impact, load ordering) would require the cllayerinfo tool which depends on this PR, but still needs some work. I propose to merge this one once the issues above have been resolved, and will implement the CI in  https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/175
@MathiasMagnus @bashbaug does that seem a reasonable path forward?

bashbaug:
Discussed in the September 5th teleconference.  Agreed to these changes in principle, and we're aiming to merge in the next couple of days.

TODOs:
- Independently test on Linux.
- Detailed code review - ideally we get one or two more approvals before merging.

Kerilk:
@MathiasMagnus , @bashbaug the rebase is done, and the CI passes. I retested on my Ubuntu (22.04) and both `OPENCL_LAYER_PATH` and the default path (`/etc/OpenCL/layers/`) behaved as expected.

loader/linux/icd_linux.c
```diff
@@ -36,111 +37,168 @@ static pthread_once_t initialized = PTHREAD_ONCE_INIT;
...
+
+struct dirElem
+{
+    char *d_name;
+    unsigned char d_type;
+};
+
+static int compareDirElem(const void *a, const void *b)
+{
+    return strcoll(((struct dirElem *)a)->d_name, ((struct dirElem *)b)->d_name);
```
bashbaug, line 131: I don't have a strong opinion about this but I do want to confirm that strcoll is intended here rather than strcmp, just because no other part of the ICD loader uses strcoll.  If it is intended, I'd suggest adding a comment to avoid confusion in the future.

bashbaug (commented):
This looks really great!  I tested layers both in the default `/etc/OpenCL/layers` and `OPENCL_LAYER_PATH` environment variable, including layers in different orders, and things worked as expected.

Would you mind bumping `OPENCL_ICD_LOADER_VERSION_REV` after this change?

I have one other question but it is not critical.

bashbaug:
Looks like this needs a rebase now also - hopefully not too bad.

MathiasMagnus:
[Windows](https://github.com/StreamHPC/OpenCL-Layers/actions/runs/3027003399) and [Linux](https://github.com/StreamHPC/OpenCL-Layers/actions/runs/3027003405) CI running.

loader/linux/icd_linux.c
Kerilk: Yes it is intended, this is the way directory entries are sorted alphanumericaly on unixes (as far as I understand such things). I wanted our sorting to behave with minimum surprise for the user. I will add a comment.

Kerilk:
> This looks really great! I tested layers both in the default `/etc/OpenCL/layers` and `OPENCL_LAYER_PATH` environment variable, including layers in different orders, and things worked as expected.
> 
> Would you mind bumping `OPENCL_ICD_LOADER_VERSION_REV` after this change?

Will do, and I will rebase.

loader/linux/icd_linux.c
Kerilk: Digging deeper, maybe I should have used libc alphasort directly, didn't know about it...

bashbaug: I see, thanks!  Adding a comment is fine for now - no need to make major changes.

Kerilk:
@bashbaug @Kerilk Thanks again for the reviews, I added the reasoning about `strcoll` usage, rebased on the latest main, and incremented windows version number.

PR #158, opened 2021-12-01
Remove const warning

keryell:
While compiling SYCL oneAPI DPC++ from https://github.com/triSYCL/sycl/tree/sycl/unified/next I got this warning:
```
[2275/2925] Building C object _deps/ocl-icd-build/CMakeFiles/OpenCL.dir/loader/linux/icd_linux.c.o
In file included from _deps/ocl-icd-src/loader/icd.h:22,
                 from _deps/ocl-icd-src/loader/linux/icd_linux.c:19:
_deps/ocl-icd-src/loader/linux/icd_linux.c: In function ‘khrIcdOsVendorsEnumerate’:
_deps/ocl-icd-src/loader/icd_platform.h:29:25: warning: initialization discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
   29 | #define ICD_VENDOR_PATH "/etc/OpenCL/vendors";
      |                         ^~~~~~~~~~~~~~~~~~~~~
_deps/ocl-icd-src/loader/linux/icd_linux.c:43:24: note: in expansion of macro ‘ICD_VENDOR_PATH’
   43 |     char* vendorPath = ICD_VENDOR_PATH;
      |                        ^~~~~~~~~~~~~~~
```
 
This PR should fix the problem.
Perhaps we could have in the CI some compilation with modern compilers and `-Werror` on Linux?

PR #159, opened 2021-12-01
Pedantic

Kerilk:
These patches allow building the loader with very strict flags on Linux using `gcc-11.2.0`.
The flags used to test it are: 
`cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS="-Wall -Werror -Wextra -pedantic" -DCMAKE_C_EXTENSIONS=OFF`
This PR is WIP and is intended as a strawman to solve the issues raised by https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/158

### DIFF
--- a/loader/linux/icd_linux.c
+++ b/loader/linux/icd_linux.c
@@ -24,6 +24,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <sys/types.h>
+#include <sys/stat.h>
 #include <dirent.h>
 #include <pthread.h>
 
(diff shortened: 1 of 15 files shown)


loader/linux/icd_linux.c
```diff
@@ -24,6 +24,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <sys/types.h>
+#include <sys/stat.h>
```
keryell, line 27: Talking about being pedantic, what about sorting the `#include`? :-)

loader/linux/icd_linux.c
```diff
@@ -61,76 +62,74 @@ void khrIcdOsVendorsEnumerate(void)
...
-            switch(dirEntry->d_type)
+            struct stat statBuff;
+            stat(dirEntry->d_name, &statBuff);
+            if(S_ISREG(statBuff.st_mode) || S_ISLNK(statBuff.st_mode))
             {
-            case DT_UNKNOWN:
-            case DT_REG:
-            case DT_LNK:
+                const char* extension = ".icd";
+                FILE *fin = NULL;
```
keryell, line 70:
```suggestion
                FILE* fin = NULL;
```

loader/linux/icd_linux.c
```diff
@@ -61,76 +62,74 @@ void khrIcdOsVendorsEnumerate(void)
         // attempt to load all files in the directory
         for (dirEntry = readdir(dir); dirEntry; dirEntry = readdir(dir) )
         {
-            switch(dirEntry->d_type)
+            struct stat statBuff;
+            stat(dirEntry->d_name, &statBuff);
+            if(S_ISREG(statBuff.st_mode) || S_ISLNK(statBuff.st_mode))
```
keryell, line 67:
```suggestion
            if (S_ISREG(statBuff.st_mode) || S_ISLNK(statBuff.st_mode))
```

loader/linux/icd_linux.c
```diff
@@ -61,76 +62,74 @@ void khrIcdOsVendorsEnumerate(void)
...
-            case DT_REG:
-            case DT_LNK:
+                const char* extension = ".icd";
+                FILE *fin = NULL;
+                char* fileName = NULL;
+                char* buffer = NULL;
+                long bufferSize = 0;
+
+                // make sure the file name ends in .icd
+                if (strlen(extension) > strlen(dirEntry->d_name) )
```
keryell, line 76: Is it in your coding style to have a space before the last `)` of an `if`?
I can see some, but not everywhere, so I do not know.

loader/linux/icd_linux.c
```diff
@@ -61,76 +62,74 @@ void khrIcdOsVendorsEnumerate(void)
...
-                        break;
-                    }
-                    // ignore a newline at the end of the file
-                    if (buffer[bufferSize-1] == '\n') buffer[bufferSize-1] = '\0';
-
-                    // load the string read from the file
-                    khrIcdVendorAdd(buffer);
+                    continue;
+                }
+                if (strcmp(dirEntry->d_name + strlen(dirEntry->d_name) - strlen(extension), extension) )
```
keryell, line 80:
```suggestion
                if (strcmp(dirEntry->d_name + strlen(dirEntry->d_name) - strlen(extension), extension))
```
?

loader/linux/icd_linux.c
```diff
@@ -61,76 +62,74 @@ void khrIcdOsVendorsEnumerate(void)
...
+                buffer = malloc(bufferSize+1);
+                if (!buffer)
+                {
+                    free(fileName);
+                    fclose(fin);
+                    continue;
+                }
+                memset(buffer, 0, bufferSize+1);
+                fseek(fin, 0, SEEK_SET);
+                if (bufferSize != (long)fread(buffer, 1, bufferSize, fin) )
```
keryell, line 113:
```suggestion
                if ((long)fread(buffer, 1, bufferSize, fin)  != bufferSize)
```
to avoid also some inattention bug the day we type `=` instead of `!=` or `==` :-)

keryell (commented):
Nice and quick cleaning!
Thanks.

loader/linux/icd_linux.c
Kerilk: Thanks for catching this one.

Kerilk:
@keryell Thanks for the review.
Unfortunately, When I changed from the switch statement to the if, I had to de-indent the whole block of code. So it feels like all of this is new, but the inner block is mostly untouched. So I indeed did not fix the styling issued you raised here.
I will maybe wait for another opinion before merging those, as I noticed the same kind of issues elsewhere and we may want to fix all of those in a different pull request.
I did commit the space I forgot in the if block though.

test/driver_stub/cl.c
```diff
@@ -902,6 +902,7 @@ clCompileProgram(cl_program            program ,
                  void (CL_CALLBACK *   pfn_notify)(cl_program  program , void *  user_data),
                  void *                user_data) CL_API_SUFFIX__VERSION_1_2
 {
+    (void)input_headers;
```
bashbaug, line 905: I have a slight preference to log `input_headers` here and in the test itself vs. marking it as unused, but I understand if you don't want to make this change right now.

bashbaug (approved):
Thanks, this is a nice improvement!

Would you like to fix the minor/cosmetic review comments or should I merge these changes as-is?

Kerilk:
Thanks for the review @bashbaug, I think that if we want to fix the style, we should agree on a clang format set of rules and do a pull request only fixing styling issues. From my implementation of the layers, I noticed similar inconsistencies to those that @keryell noticed and I 'd rather fix all of them at the same time in a consistent manner rather than dilute those among several patches.

test/driver_stub/cl.c
Kerilk: I had the same though as you at first, but then I started wondering if the ignored parameters were not dynamically allocated buffers that would add randomness into the trace and render the comparison to the reference difficult. I agree it would be nice checking them as well, but it may require some changes to the infrastructure.

bashbaug:
Makes sense to me, I'm fine merging these changes as-is.  Thanks again!

loader/linux/icd_linux.c
```diff
@@ -61,76 +62,74 @@ void khrIcdOsVendorsEnumerate(void)
         // attempt to load all files in the directory
         for (dirEntry = readdir(dir); dirEntry; dirEntry = readdir(dir) )
         {
-            switch(dirEntry->d_type)
+            struct stat statBuff;
+            stat(dirEntry->d_name, &statBuff);
```
jrprice, line 66: This PR has broken our internal tests, because this `stat` call is always failing with "No such file or directory".

I think this needs to be the full absolute path (e.g. the `fileName` string that is constructed below), instead of the relative `dirEntry->d_name` path used here.

Kerilk: You are absolutely correct, I made a fix pull request here: https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/161
Thanks for finding this.

PR #161, opened 2021-12-16
Fix: stat was done on an incomplete file path.

Kerilk:
This should fix the issue raised by @jrprice.
Sorry about that, I don't see the issue anymore.


bashbaug (approved):
Thank you for the quick fix.  I've also verified that this fixes the issue.

jrprice:
Verified that this fixes it for me too. Thanks for the quick fix!

Kerilk:
Thanks for the quick reviews :)

PR #163, opened 2022-01-21
Renamed BUILD_SHARED_LIBS option

LeonBrands:
Changed BUILD_SHARED_LIBS to OPENCL_ICD_LOADER_BUILD_SHARED_LIBS, so that the option doesn't change the default library type. Changing BUILD_SHARED_LIBS currently affects all other libraries in the CMake build when the loader is added using add_subdirectory.


bashbaug:
Hi @LeonBrands, thank you for your contribution!

Can you say a bit more what you are trying to do?  By any chance, are you trying to compile a statically linked ICD loader but without changing all of your other libraries to static libraries?

For some of my own projects I've been adding controls to force a static lib but to follow the general cmake BUILD_SHARED_LIBS otherwise, see for example:

* https://github.com/bashbaug/opencl-extension-loader/blob/main/CMakeLists.txt#L87

Would this work for your use-case?  If so, would this be a better solution for the ICD loader?

CC @MathiasMagnus in case he knows of a known cmake best practice to handle things like this also.

LeonBrands:
Hi! Thank you for your response,

No I'm trying to build the loader with shared libs, but as a part of a greater CMake project. The problem I'm having has to do with the line where it declares BUILD_SHARED_LIBS as an option, defaulted to true. This isn't a problem for my own libraries (I use STATIC and SHARED for all add_library calls), but other external dependencies in my project do not.

Because the loader sets BUILD_SHARED_LIBS to true, these other dependencies in my project start unexpectedly compiling as shared when I don't intend them to. My current workaround is that I forcefully set BUILD_SHARED_LIBS to OFF again after I add OpenCL, but I'd prefer if it didn't touch global state at all

MathiasMagnus:
@LeonBrands thanks for your submission. I played around with `BUILD_SHARED_LIBS` to see what impact changing it would have on existing projects and ones such as yours.

IMHO projects making the building other projects as part of their own build is fairly new practice. 3-4 years ago even tools like VSCode CMake Tools threw a warning for having multiple project statements in one build. Not all projects were designed to play nice with this type of behavior.

I'll give a proper response on Monday, but I think we could incorporate your change, even though it's slightly breaking. Rationale with alternatives to follow.

LeonBrands:
Yeah I understand that including multiple projects in a cmake build isn't the most traditional approach, but we prefer it because by doing so we avoid having to manually build & update dependencies for every supported platform

CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 # may be useful to simplify the build and distribution of the application. E.g.
 # in test farms, or in cases where the end-user system configs are known in
 # advance. Use it with discretion.
-option (BUILD_SHARED_LIBS "Build shared libs" ON)
+option (OPENCL_ICD_LOADER_BUILD_SHARED_LIBS "Build shared libs" ON)
```
MathiasMagnus, line 44: I've played around with `BUILD_SHARED_LIBS` in a subproject scenario and indeed they don't mix well (at least when both the calling project the the called project wish to control it. Whomever gets there first (because it's a `CACHE` variable) wins (without `FORCE` that is).

As far as the grand scheme of things go the OpenCL-ICD-Loader can only guarantee build script API compatibility with itself. If a project involves building another into their own, they take on following all the changes the project makes. (It's simply impossible for a project to be backward compatible with _every_ project recursing into.) Having that said...

**For OpenCL-ICD-Loader to control this global property manually is a bug.** If a calling project favors static libs, and before us calls `option(BUILD_SHARED_LIBS "Don't touch" OFF)` the ICD-Loader by default will be built as a static library. The ICD Loader's preference to favor dynamic libraries is _indifferent to what our caller prefers or asked for_. Therefor breaking those who were in the shoes of @LeonBrands, who had to correct global state after recursing into our folder is not a break but a fix to a latent bug.

Suggestion is the following:

```suggestion
include(CMakeDependentOption)
if(DEFINED BUILD_SHARED_LIBS)
  set(OPENCL_ICD_LOADER_BUILD_SHARED_LIBS_DEFAULT ${BUILD_SHARED_LIBS})
else()
  set(OPENCL_ICD_LOADER_BUILD_SHARED_LIBS_DEFAULT ON)
endif()
  option(OPENCL_ICD_LOADER_BUILD_SHARED_LIBS "Build OpenCL ICD Loader as shared library" ${OPENCL_ICD_LOADER_BUILD_SHARED_LIBS_DEFAULT})
```

This way:
- People who consumed the ICD Loader as a separate install
  - If they don't ask for anything, they get the same `SHARED` as they used to.
  - If they ask for `SHARED` or `STATIC` like they used to using `BUILD_SHARED_LIBS`, they get the lib type they expect.
- People who consume the ICD by making the build part of theirs
  - If they set `BUILD_SHARED_LIBS` to `ON` or `OFF`, the loader is built as `SHARED`/`STATIC` as the global setting implies.
  - If they don't set anything, _BUGFIX_ the new option when leaking won't alter the global default (which is building `STATIC` when `BUILD_SHARED_LIBS` isn't set to anything).
  - If they wish to mix and match _NEW FEATURE_ only then and only they have to opt in via the new variable `OPENCL_ICD_LOADER_BUILD_SHARED_LIBS`.

This way people get the escape hatch they want but existing practice will not change.

_(ps.: the reason why this code is ugly is because unlike most variables, `BUILD_SHARED_LIBS` has no default value, the default isn't `OFF`, but it's absence implies `OFF` too, therefor the need for `OPENCL_ICD_LOADER_BUILD_SHARED_LIBS_DEFAULT`.)_

CMakeLists.txt
```diff
@@ -93,8 +93,14 @@ endif ()
...
 
-add_library (OpenCL ${OPENCL_ICD_LOADER_SOURCES})
-add_library (OpenCL::OpenCL ALIAS OpenCL)
+if (${OPENCL_ICD_LOADER_BUILD_SHARED_LIBS} OR ${BUILD_SHARED_LIBS})
+  add_library (OpenCL SHARED ${OPENCL_ICD_LOADER_SOURCES})
+  add_library (OpenCL::OpenCL ALIAS OpenCL)
+else()
+  add_library (OpenCL SHARED ${OPENCL_ICD_LOADER_SOURCES})
+  add_library (OpenCL::OpenCL ALIAS OpenCL)
+endif()
```
MathiasMagnus, line 102: `add_library (OpenCL::OpenCL ALIAS OpenCL)` need not be duplicated. It can go outside the conditional.

MathiasMagnus (changes requested):
This is my suggestion, @bashbaug let me know if this seems reasonable.

CMakeLists.txt
LeonBrands: Thank you for your response! This is a good solution that preserves the global state and OpenCL's preference for dynamic loaders 👍

CMakeLists.txt
LeonBrands: Done

CMakeLists.txt
```diff
@@ -93,8 +99,14 @@ endif ()
 
 set (OPENCL_ICD_LOADER_HEADERS_DIR "${CMAKE_CURRENT_SOURCE_DIR}/inc" CACHE PATH "Path to OpenCL Headers")
 
-add_library (OpenCL ${OPENCL_ICD_LOADER_SOURCES})
+if (${OPENCL_ICD_LOADER_BUILD_SHARED_LIBS} OR ${BUILD_SHARED_LIBS})
```
bashbaug, line 102: Can you please check that this does what you want it to do?

Since we have already incorporated the state of `BUILD_SHARED_LIBS` into the default state for `OPENCL_ICD_LOADER_BUILD_SHARED_LIBS` above I think we might just want this to be:

```cmake
if (${OPENCL_ICD_LOADER_BUILD_SHARED_LIBS})
```

CMakeLists.txt
```diff
@@ -93,8 +99,14 @@ endif ()
...
+  add_library (OpenCL SHARED ${OPENCL_ICD_LOADER_SOURCES})
+else()
+  add_library (OpenCL SHARED ${OPENCL_ICD_LOADER_SOURCES})
```
bashbaug, line 105: This looks like a typo because it matches the "if" block above.  Should this be:

```suggestion
  add_library (OpenCL STATIC ${OPENCL_ICD_LOADER_SOURCES})
```

bashbaug (commented):
Thanks, the high-level idea seems good.  Just a few details to check...

CMakeLists.txt
LeonBrands: Right, this can be removed now that the suggestion from Mathias is in place

CMakeLists.txt
```diff
@@ -41,7 +41,13 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 # may be useful to simplify the build and distribution of the application. E.g.
 # in test farms, or in cases where the end-user system configs are known in
 # advance. Use it with discretion.
-option (BUILD_SHARED_LIBS "Build shared libs" ON)
+include(CMakeDependentOption)
```
bashbaug, line 44: Minor: I don't think this PR introduces any dependent options so perhaps this can be removed?

bashbaug (approved):
Thanks, latest updates LGTM!

CMakeLists.txt
LeonBrands: Yeah this can go

bashbaug:
@MathiasMagnus any idea what's going on with the CI builds?

I'm not too worried about the AppVeyor failures (though we need to disable that so it stops popping up), but I do want the GitHub actions to be clean and I don't think the failures are caused by the changes in this PR.

MathiasMagnus:
I think they are caused by this PR, because there are more libraries built than just the ICD Loader itself, and now the rest of the print and ICD stubs became static libs. Those libs need the LIB_TYPE addition like the ICD itself.

LeonBrands:
I believe I've tracked down all the subdirectories that expected BUILD_SHARED_LIBS to be set, the tests should succeed now

MathiasMagnus (approved):

MathiasMagnus:
@LeonBrands I may have been unclear. The ICD Loader loads implementations (ICDs, Installable Client Drivers) using [`dlopen()`](https://linux.die.net/man/3/dlopen) and equivalent functions. As the name `dlopen` suggests it can only open dynamic libraries. The Loader itself can be a DLL or it can be statically linked to an application (read the CMake comment preceding the option to see when this is highly unrecommended). However, the things it loads _have to_ be dynamic libraries.

The tests previously didn't fail because we don't build the ICD loader as a static lib in CI ever (perhaps we should change that).

The Windows tests pass because they are bugged (they were recently updated and they only fail if the last command fails, however that's a registry modification, so that always succeedes. Currently tests are passing, but because we don't build the ICD (and consequently the mock ICDs) as static.

The solution is to build the 3 targets you modfiied in your last commit (`OpenCLDriverStub`, `PrintLayer`, `IcdLog`) to be built as `SHARED` unconditionally. They only ever make sense that way.

MathiasMagnus (changes requested):
@LeonBrands I may have been unclear. The ICD Loader loads implementations (ICDs, Installable Client Drivers) using [`dlopen()`](https://linux.die.net/man/3/dlopen) and equivalent functions. As the name `dlopen` suggests it can only open dynamic libraries. The Loader itself can be a DLL or it can be statically linked to an application (read the CMake comment preceding the option to see when this is highly unrecommended). However, the things it loads _have to_ be dynamic libraries.

The tests previously didn't fail because we don't build the ICD loader as a static lib in CI ever (perhaps we should change that).

The Windows tests pass because they are bugged (they were recently updated and they only fail if the last command fails, however that's a registry modification, so that always succeedes. Currently tests are passing, but because we don't build the ICD (and consequently the mock ICDs) as static.

The solution is to build the 3 targets you modfiied in your last commit (`OpenCLDriverStub`, `PrintLayer`, `IcdLog`) to be built as `SHARED` unconditionally. They only ever make sense that way.

LeonBrands:
I see! In the previous version they would've built as static if BUILD_SHARED_LIBS was set to OFF so I figured I'd replicate that behaviour. I'll change them to just shared

MathiasMagnus (approved):
Super, thank you for your contribution.

Issue #167, opened 2022-03-16
Enable the internal trace in ICD loader


cdai2:
Some times the OpenCL ICDs are not loaded successfully because of some reason.  But the user can not know the reason quickly. It's better that OpenCL ICD loader can provide some trance information or trace information.

Currently there is "KHR_ICD_TRACE" to provides some trance info but it's not enabled by default.  It's commented out by "#if 0" in icd.h fie.  I propose to use one environment variable to enable this "KHR_ICD_TRACE" macro so that user can get trace information.

Kerilk:
This is a common suggestion, thanks for putting it back to the top of the list, and I think it is reasonable. Before we start we need to decide:
 - Name of the environment variable
 - Stream to use (configurable?)
 - Should we provide a registry key providing the same functionality as environment variables cannot be used on Windows while running securely
Are there other constraints/features I am overlooking?

cdai2:
Thanks for feedback. Currently I have a draft patch.  I will submit for review soon:

- Name of the environment variable
      I used "OCL_ICD_ENABLE_TRACE" as environment variable following the pattern of existing ICD environment variables.

- Stream to use (configurable?)
      I use  "stderr".  We can add configurable stream.

- Should we provide a registry key providing the same functionality as environment variables cannot be used on Windows while running securely
     Personally I think the trace information is not secure.  The environment variable should be accessible under securely execution environment.  We do not need to provide registry key for to provide same functionality as environment variable.

cdai2:
Please review https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/168.

bashbaug:
I think this is fixed now, by #168?  Closing, but please feel free to reopen if there is still work to do, thanks!

PR #168, opened 2022-03-30
Enable trace by OCL_ICD_ENABLE_TRACE environment variable

cdai2:


Kerilk:
@jenatali Would you be able to take a look?
This impacts Windows too (see https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/167).

loader/icd.c
```diff
@@ -26,13 +26,16 @@
 #include <string.h>
 
 KHRicdVendor *khrIcdVendors = NULL;
+char *khrEnableTrace = NULL;
```
Kerilk, line 29: I would use an integer here to store the state of the tracing activation.
The string returned by getenv should not change through the execution, so better compare it early.

loader/icd.c
```diff
@@ -26,13 +26,16 @@
...
+
 #if defined(CL_ENABLE_LAYERS)
 struct KHRLayer *khrFirstLayer = NULL;
 #endif // defined(CL_ENABLE_LAYERS)
 
 // entrypoint to initialize the ICD and add all vendors
 void khrIcdInitialize(void)
 {
+    khrEnableTrace = khrIcd_getenv("OCL_ICD_ENABLE_TRACE");
```
Kerilk, line 38: Two remarks here

- Same remark as in the issue, discussion of using `khrIcd_getenv` vs `khrIcd_secure_getenv`. Tagging @jenatali as the windows implications seem stronger than the Linux ones (I may be wrong).
- the initialization of the tracing variable should go in it's own subroutine, maybe per os, depending on the result of the above discussion. This initialization should end up with `khrEnableTrace` containing a boolean value signifying if the tracing is active or not. The list of acceptable values for the string will need to be determined but will most probably end being more than just `"True"`

loader/icd.h
```diff
@@ -83,6 +84,7 @@ struct KHRicdVendorRec
 
 // the global state
 extern KHRicdVendor * khrIcdVendors;
+extern char *khrEnableTrace;
```
Kerilk, line 87: see above comment regarding variable type.

loader/icd.h
```diff
@@ -152,51 +154,58 @@ void khrIcdContextPropertiesGetPlatform(
...
+#define KHR_ICD_TRACE(...) \
+if (khrEnableTrace && strcmp(khrEnableTrace, "True") == 0) \
+{ \
     do \
     { \
         fprintf(stderr, "KHR ICD trace at %s:%d: ", __FILE__, __LINE__); \
         fprintf(stderr, __VA_ARGS__); \
-    } while (0)
+    } while (0); \
+}
```
Kerilk, line 165: I would rather see something like this:
```C
if (khrEnableTrace) \
    do \
    {
        fprintf(stderr, "KHR ICD trace at %s:%d: ", __FILE__, __LINE__); \
        fprintf(stderr, __VA_ARGS__); \
    } while (0)
```
As this will keep the established usage of requiring a `;` after each macro invocation. I also simplified the test to reflect the usage of the boolean/int variable.

loader/icd.h
```diff
@@ -152,51 +154,58 @@ void khrIcdContextPropertiesGetPlatform(
...
         } \
-    } while (0)
-#else
-    #define KHR_ICD_TRACE(...)
-    #define KHR_ICD_WIDE_TRACE(...)
-    #define KHR_ICD_ASSERT(x)
-#endif
-
+    } while (0); \
+}
```
Kerilk, line 191: Same as above.

loader/icd.h
```diff
@@ -152,51 +154,58 @@ void khrIcdContextPropertiesGetPlatform(
...
     do \
     { \
         if (!handle) \
         { \
             return invalid_handle_error_code; \
         } \
-    } while (0)
-
+    } while (0); \
+}
```
Kerilk, line 203: Same as above.

loader/icd.h
```diff
@@ -207,8 +216,7 @@ void khrIcdContextPropertiesGetPlatform(
             } \
             return NULL; \
         } \
-    } while (0)
-
+    } while (0); \
+}
```
Kerilk, line 220: Same as above.

loader/linux/icd_linux.c
```diff
@@ -169,7 +169,12 @@ void khrIcdOsVendorsEnumerateOnce(void)
 // dynamically load a library.  returns NULL on failure
 void *khrIcdOsLibraryLoad(const char *libraryName)
 {
-    return dlopen (libraryName, RTLD_NOW);
+    void* ret = dlopen (libraryName, RTLD_NOW);
+    if (NULL == ret)
+    {
+        KHR_ICD_TRACE("Failed to load driver because %s.", dlerror());
```
Kerilk, line 175:
```suggestion
        KHR_ICD_TRACE("Failed to load driver because %s.\n", dlerror());
```

Kerilk (changes requested):
Thanks for the PR, it is greatly appreciated.
I do have some comments/change requests, and also requested others review those changes.

loader/linux/icd_linux.c
cdai2: Fixed.

loader/icd.h
cdai2: Fixed

loader/icd.h
cdai2: Fixed

loader/icd.h
cdai2: Fixed

loader/icd.h
cdai2: Fixed

loader/icd.h
cdai2: Fixed

loader/icd.c
cdai2: Fixed. I defined it as integer.

loader/icd.c
cdai2: For item 1,  I will wait for @jenatali 's feedback.  Personally I still think this env variable does not change the execution logic of loading drivers. So we do not need the khrIcd_secure_getenv.

For item 2, I added "true", "T", "1".   I added the initialization of trace enabling because khrIcdInitialize is the entry point to initialize ICD.

cdai2:
@Kerilk Thanks for your review.  I fixed most of issues according to your comments.

loader/icd_dispatch.c
```diff
@@ -17,6 +17,7 @@
  */
 
 #include "icd_dispatch.h"
+#include "icd_envvars.h"
```
Kerilk, line 20: This file should be reverted.
```suggestion
```

loader/icd_dispatch_generated.c
```diff
@@ -17,8 +17,11 @@
  */
 
 #include "icd_dispatch.h"
+#include "icd_envvars.h"
 #include "icd.h"
 
+#include <string.h>
+
 #ifdef __cplusplus
```
Kerilk, line 25: This file should be reverted.
```suggestion
#include "icd_dispatch.h"
#include "icd.h"

#ifdef __cplusplus
```

loader/windows/icd_windows_apppackage.cpp
```diff
@@ -19,6 +19,10 @@
 #include "icd.h"
 #include "icd_windows.h"
 
+extern "C" {
+    int khrEnableTrace;
+}
+
```
Kerilk, line 25: Not sure about this one (don't know why it would be required), but it would seem more coherent with below conventions if it were written as:

```suggestion
extern "C" int khrEnableTrace;

```

loader/icd.c
```diff
@@ -26,13 +26,24 @@
...
 {
+    char *enableTrace = khrIcd_getenv("OCL_ICD_ENABLE_TRACE");
+    if (enableTrace && (strcmp(enableTrace, "True") == 0 ||
+            strcmp(enableTrace, "true") == 0 ||
+            strcmp(enableTrace, "T") == 0 ||
+            strcmp(enableTrace, "1") == 0))
+    {
+	khrEnableTrace = 1;
+    }
+
```
Kerilk, line 46: This should go in its own subroutine. also, line 44 was indented with a tabulation.

Kerilk (changes requested):
Thanks for the update. Another round of comments on those changes, and some I missed the first time around.

loader/icd.h
```diff
@@ -37,6 +37,7 @@
 #include <CL/cl_ext.h>
 #include <CL/cl_icd.h>
 
+#include <stdio.h>
```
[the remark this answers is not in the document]
cdai2: fixed.

loader/icd.h
```diff
@@ -152,51 +154,52 @@ void khrIcdContextPropertiesGetPlatform(
...
-    #include <stdio.h>
-    #define KHR_ICD_TRACE(...) \
+#define KHR_ICD_TRACE(...) \
+if (khrEnableTrace) \
     do \
     { \
         fprintf(stderr, "KHR ICD trace at %s:%d: ", __FILE__, __LINE__); \
         fprintf(stderr, __VA_ARGS__); \
-    } while (0)
+    } while (0);
```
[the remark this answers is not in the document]
cdai2: fixed

loader/icd.h
```diff
@@ -152,51 +154,52 @@ void khrIcdContextPropertiesGetPlatform(
...
         { \
             fprintf(stderr, "KHR ICD assert at %s:%d: %s failed", __FILE__, __LINE__, #x); \
         } \
-    } while (0)
-#else
-    #define KHR_ICD_TRACE(...)
-    #define KHR_ICD_WIDE_TRACE(...)
-    #define KHR_ICD_ASSERT(x)
-#endif
+    } while (0);
```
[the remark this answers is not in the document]
cdai2: fixed.

loader/icd.h
```diff
@@ -207,8 +210,6 @@ void khrIcdContextPropertiesGetPlatform(
             } \
             return NULL; \
         } \
-    } while (0)
-
+    } while (0);
```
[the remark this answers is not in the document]
cdai2: fixed

loader/icd.h
```diff
@@ -152,51 +154,52 @@ void khrIcdContextPropertiesGetPlatform(
...
+
 #ifdef _WIN32
 #define KHR_ICD_WIDE_TRACE(...) \
+if (khrEnableTrace) \
     do \
     { \
         fwprintf(stderr, L"KHR ICD trace at %hs:%d: ", __FILE__, __LINE__); \
         fwprintf(stderr, __VA_ARGS__); \
-    } while (0)
+    } while (0);
```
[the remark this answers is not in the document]
cdai2: fixed.

loader/icd_dispatch.c
cdai2: fixed

loader/icd_dispatch_generated.c
cdai2: fixed

loader/icd.c
cdai2: I added on function named khrIcdInitializeTrace for it.

cdai2:
Please review the fix again.  thanks @Kerilk

loader/windows/icd_windows_apppackage.cpp
Kerilk: Leaving this open so @jenatali can take a look.

loader/icd_dispatch_generated.c
```diff
@@ -19,6 +19,8 @@
 #include "icd_dispatch.h"
 #include "icd.h"
 
+#include <string.h>
+
 #ifdef __cplusplus
```
Kerilk, line 24: This file is generated, so it shouldn't be modified. Also I don't understand why this include would suddenly be required.
If the modification was required, https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/main/scripts/icd_dispatch_generated.c.mako would need to be updated before regenerating the dispatch.
```suggestion
#ifdef __cplusplus
```

Kerilk (changes requested):
Thanks for the changes. The generated dispatch was only partially reverted, but aside from that looks good to me. @bashbaug and @jenatali can perhaps take a look and maybe spot what I missed.

loader/icd_dispatch_generated.c
cdai2: Fixed.  thanks

cdai2:
@bashbaug  Please help to review.

README.md
```diff
@@ -146,3 +146,4 @@ The following debug environment variables are available for use with the OpenCL
 | OCL_ICD_FILENAMES                 | Specifies a list of additional ICDs to load.  The ICDs will be enumerated first, before any ICDs discovered via default mechanisms. | `export OCL_ICD_FILENAMES=libVendorA.so:libVendorB.so`<br/><br/>`set OCL_ICD_FILENAMES=vendor_a.dll;vendor_b.dll` |
 | OCL_ICD_VENDORS                   | On Linux and Android, specifies a directory to scan for ICDs to enumerate in place of the default `/etc/OpenCL/vendors'. |  `export OCL_ICD_VENDORS=/my/local/icd/search/path` |
 | OPENCL_LAYERS                    | Specifies a list of layers to load. |  `export OPENCL_LAYERS=libLayerA.so:libLayerB.so`<br/><br/>`set OPENCL_LAYERS=libLayerA.dll;libLayerB.dll` |
+| OCL_ICD_ENABLE_TRACE             | Enable the trace machanism          |  `export OCL_ICD_ENABLE_TRACE=True`<br/>`true, T, 1 can also be used here.`  |
```
bashbaug, line 149:
```suggestion
| OCL_ICD_ENABLE_TRACE             | Enable the trace mechanism          |  `export OCL_ICD_ENABLE_TRACE=True`<br/>`true, T, 1 can also be used here.`  |
```

loader/icd.h
```diff
@@ -152,40 +154,40 @@ void khrIcdContextPropertiesGetPlatform(
     cl_platform_id *outPlatform);
 
 // internal tracing macros
-#if 0
-    #include <stdio.h>
-    #define KHR_ICD_TRACE(...) \
+#define KHR_ICD_TRACE(...) \
+if (khrEnableTrace) \
```
bashbaug, line 158: This new if check should be moved inside of the one-trip do-while loop, rather than be placed outside.

For details, see: https://wiki.sei.cmu.edu/confluence/display/c/PRE10-C.+Wrap+multistatement+macros+in+a+do-while+loop

Applies here and the other places in this file where an if check was added.

bashbaug (changes requested):
Mostly LGTM, just a few minor issues to fix.  This will be very useful!

I think we should also bump the Windows revision in OpenLC.rc after these changes?  This will make it easier to identify ICD loaders with the trace functionality vs. without.

Thanks!

bashbaug (commented):
I am on vacation next week and will not be checking GitHub regularly.  I'm fine merging these changes as soon as my comments above are resolved.  Thank you!

cdai2:
Thanks a lot for you to review.  I will update it immediately.

cdai2:
> 

Which number should be bumped in OpenCL.rc?  REV or MAJOR or MINOR ? 

#define OPENCL_ICD_LOADER_VERSION_MAJOR 3
#define OPENCL_ICD_LOADER_VERSION_MINOR 0
#define OPENCL_ICD_LOADER_VERSION_REV   1

cdai2:
@bashbaug   Please review. 
 I updated it according to your comments.  The OPENCL_ICD_LOADER_VERSION_REV  in OpenCL.rc is increased from 1 to 2.

Kerilk (approved):

README.md
cdai2: done

cdai2:
Ping

cdai2:
Can anyone help me to check or guide me to check the check error "Windows / msvc (v141, ON, Visual Studio 17 2022, x64, 90, 3.22.0) (pull_request) Cancelled after 2m — msvc (v141, ON, Visual Studio 17 2022, x64, 90, 3.22.0)"？   

I cannot reproduce any compilation error on my local machine with visual studio 2017.

Kerilk:
I think you can safely ignore those, they are independent from your modifications. I'll let @bashbaug hit the merge button when he is satisfied that his requests have been incorporated.

loader/windows/icd_windows_apppackage.cpp
jenatali: Weird. No idea why it'd do that.

Alternatively you could use:
```c++
#ifdef __cplusplus
extern "C" {
#endif

...

#ifdef __cplusplus
}
#endif
```

In the `icd.h` header to avoid having to re-declare it here with the extern.

jenatali (approved):
Looks good to me.

bashbaug (approved):
Latest changes LGTM - thanks!

bashbaug:
There are three approvals for this PR now so I think we can safely merge this.

Note, the CI failures should be fixed by https://github.com/KhronosGroup/OpenCL-Headers/pull/198.

PR #169, opened 2022-04-12
Install PDBs

MathiasMagnus:
Even though console tracing is soon merged (and it's mighty useful), in some cases, for eg. when one is writing an ICD from scratch, debugging dangling pointers, mismatching levels of pointer indirection, lifetime issues, etc. attaching a debugger to the ICD is necessary. In such cases, it's useful if one can install a debug build of the ICD (typically to a non-system location) for debugging purposes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,11 @@ install(
   LIBRARY
     DESTINATION ${CMAKE_INSTALL_LIBDIR} # obtained from GNUInstallDirs
 )
+install(
+  FILES $<TARGET_PDB_FILE:OpenCL>
+  DESTINATION ${CMAKE_INSTALL_BINDIR}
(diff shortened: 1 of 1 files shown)


bashbaug:
I tried this PR locally and I see PDBs being built but not installed.  Do I need to do anything special to enable this?

MathiasMagnus:
@bashbaug I fixed it. I fell victim to the "looks OK, but..." category. Even though dedicated genexprs and properties exist in CMake, none of them work with non-MSVC toolchains. This works with `cl.exe/link.exe` , `clang-cl.exe/lld-link.exe` and `clang.exe/lld.exe` .

bashbaug (approved):
Thanks!  I've verified that PDBs are copied with these latest changes.

PR #171, opened 2022-04-25
Fix Windows compilation error for Trace mechanism

cdai2:
There is compilation error for some compiler because the global variable
in header file is included by both c and cpp file.

### DIFF
--- a/loader/icd.h
+++ b/loader/icd.h
@@ -84,7 +84,14 @@ struct KHRicdVendorRec
 
 // the global state
 extern KHRicdVendor * khrIcdVendors;
-extern int khrEnableTrace;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+    extern int khrEnableTrace;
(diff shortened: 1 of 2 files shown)


cdai2:
Ping @jenatali.  Please review and approve.  I did not notice your previous comments in #168.  It should be implemented as your suggestion. I fixed here.

jenatali (approved):

cdai2:
ping @bashbaug @Kerilk   Please help to approve and land.  It blocks our project's pull down.  
Sorry for effort introduced by my previous incorrect patch.

Kerilk (approved):
Looks good.

The while thing looks a bit brittle.
In the future, if we need to convert part of the project to C++, we should have the headers fully protected for C++ inclusion.

bashbaug (commented):
LGTM

Can we add the compilers that were failing to our CI testing?

PR #177, opened 2022-05-24
Build ICD Loader using Clang on Windows

MathiasMagnus:
This work was born out of grievance of reproducing non-MSVC specific warnings, errors and linker failures on the OpenCL-Layers project _on Windows_ without having to fire up containers or building-installing under WSL. (At a larger scale, using Clang-derivates on Windows such as ComputeCpp often blow up the console with warnings if the device compiler sees the OpenCL headers.)

Beside touching up the source code to fix warnings, two major things were implemented in CI:
- Extend Windows CI coverage to using `clang-cl.exe`, the MSVC-like CLI of Clang.
  - This was done while porting CI definition to PowerShell and doing better error checking
  - Improving error checking uncovered a lot of shortcomings of the previous CI definition (lots of errors went unnoticed)
- Simplify Linux CI definition and accelerate it by using the same build image as the SDK does

_Note: Overall I'm fairly unsatisfied how much code getting this range of CI coverage takes. (Linux/Mac/Win, GCC/Clang/MSVC, C99/11/17, with/without lang exts, Debug/Release) CI is doing the very same thing, but with 2-3 lines of difference for every test matrix elem. I'll be thinking about how to achieve the same coverage with a unified script and much less duplication of code. I would like the scripts to be **much** simpler._

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -3,20 +3,32 @@ name: Linux
 on: [push, pull_request]
 
 jobs:
-  cmake-minimum:
-    runs-on: ${{ matrix.OS }}
+  compatibility:
+    runs-on: ubuntu-18.04
+    container: streamhpc/opencl-sdk-intelcpu:ubuntu-18.04-20220127
     strategy:
       matrix:
-        OS: [ubuntu-18.04]
-        COMPILER: [gcc-7, gcc-8, clang-8, clang-10]
-        EXT: [ON, OFF]
+        C_COMPILER: [gcc-7, gcc-11, clang-8, clang-13]
+        CMAKE: [3.1.3]
         GEN: [Unix Makefiles]
-        CONFIG: [Debug, Release]
+        CONFIG: [Release]
+        BIN: [64, 32]
         STD: [99, 11]
-        BIN: [32, 64]
-        CMAKE: [3.1.3, 3.21.2]
+        EXT: [ON, OFF]
+        include:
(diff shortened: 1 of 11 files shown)


.github/workflows/linux.yml
```diff
@@ -31,194 +43,61 @@ jobs:
...
-
-
-
-
-
-  cmake-latest:
-    runs-on: ${{ matrix.OS }}
-    strategy:
-      matrix:
-        OS : [ubuntu-20.04]
```
Kerilk, line 143: Ubuntu 20.04 is not tested anymore after those changes. Is this intended?

loader/windows/icd_windows.c
```diff
@@ -279,7 +282,19 @@ void *khrIcdOsLibraryGetFunctionAddress(void *library, const char *functionName)
...
+#elif defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wpedantic"
+#endif
     return GetProcAddress( (HMODULE)library, functionName);
+#if defined(_MSC_VER) && !defined(__clang__)
+#pragma warning( pop )
+#elif defined(__clang__)
+#pragma clang diagnostic pop
+#endif
```
Kerilk, line 297: I would try using:
```suggestion
    return (void *)(intptr_t)GetProcAddress( (HMODULE)library, functionName);
```
here. I would be consistent with what we do at other place in the code. You may require an additional include.

PR #179, opened 2022-06-23
Rewrite CLOn12 loader in C

jenatali:
I learned this last week that there exist C Win32 APIs for Windows package management, under the AppModel.h header. I wish I knew this when I first wrote this loader code. The C APIs are all around better for low-level code that doesn't need the WinRT language projections:
* Simpler
* Doesn't need COM/WinRT initialization - we've encountered problems where the COM initialization unintentionally happens under the Windows DLL loader lock, which fails catastrophically
* Fewer memory allocations

And best of all for this project, it means it can be a C file and aligns better with the rest of the C codebase.

I'd explored removing this entirely, and forcing our AppX package to write the normal ICD registry keys, since there are AppX extensions for writing to the registry, but those extensions are unable to expand out a path into the installed package directory. So, at least for now, this still needs to be a special case, but at least it can be simpler.


Kerilk:
That is great, thanks a lot for this contribution!

Kerilk (commented):
Disclaimer, I couldn't test this.
But from a loader developer perspective this is great. I couldn't find any issue in the code.

Maybe we can also remove the cplusplus guards around:
https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/7072cf2ae9d9acb6be8f4fc956f1cdbf0675695b/loader/icd.h#L88-L94

jenatali:
> Disclaimer, I couldn't test this.

That's okay, I was able to, and confirmed that it works the same as the old code.

> Maybe we can also remove the cplusplus guards around:
> 
> https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/7072cf2ae9d9acb6be8f4fc956f1cdbf0675695b/loader/icd.h#L88-L94

Yep, we should be able to. Want me to add that to this PR?

Kerilk:
> Yep, we should be able to. Want me to add that to this PR?

Yes, please do.

Kerilk (approved):

jenatali:
Is there anything else needed here before this can merge?

Kerilk:
@bashbaug I was looking at PR to merge for the release and this seems a good one as well.

bashbaug:
Thanks, I missed this one while I was gone.  Added it to my TODO list.

loader/windows/icd_windows_apppackage.c
```diff
@@ -0,0 +1,99 @@
...
+    {
+        KHR_ICD_TRACE("Failed to allocate memory for package names\n");
+        goto cleanup;
+    }
+
+    if (ERROR_SUCCESS != GetPackagesByPackageFamily(familyName,
+                                                    &numPackages, packages,
+                                                    &bufferLength, buffer))
+    {
+        KHR_ICD_TRACE("Failed to mapping layers package full names\n");
```
bashbaug, line 60: Not entirely sure if this is the right fix, please check.

```suggestion
        KHR_ICD_TRACE("Failed to get mapping layers package full names\n");
```

bashbaug (approved):
This is a nice simplification - thank you!

I found one minor typo (I think) in one of the tracing messages but this is fine to merge as-is.

PR #180, opened 2022-06-24
Limit calls to khrIcdInitializeTrace

Kerilk:
I realized the trace initialization function was not called once, but multiple times. This moves the function to only be used once.

### DIFF
--- a/loader/icd.c
+++ b/loader/icd.c
@@ -48,7 +48,6 @@ void khrIcdInitializeTrace(void)
 // entrypoint to initialize the ICD and add all vendors
 void khrIcdInitialize(void)
 {
-    khrIcdInitializeTrace();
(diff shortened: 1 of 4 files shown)


Kerilk:
@bashbaug requested your review as you reviewed https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/168.
Couldn't add @jenatali as reviewer, but he did review the original one as well.

jenatali (approved):

Kerilk:
@jenatali Thanks for the review!

bashbaug (approved):
LGTM.  We shouldn't be calling `khrIcdInitialize` terribly often anymore, but regardless, we do not need to call `KhrIcdInitializeTrace` more than once so this is a nice improvement.  Thanks!

PR #181, opened 2022-07-12
Prevent ICD from loading outdated amdocl binary 1.Keep track of libra…

lttamd:
Prevent ICD from loading outdated amdocl  binary 
1.Keep track of library name in KHRicdVendorRec 
2.Check if library with the same name has already been loaded before adding it in ICD list

There are going to be two sets of amdocl in registry for ICD to load if we install a new driver on top of a previously installed old driver.
It will cause subsequent undefined behaviors such as duplicate devices.

This is because AMD driver installation switched from c-inf to u-inf which no longer uses SOFTWARE\\Khronos\\OpenCL\\Vendors. 
If a new diver is installed by either user or Windows update without manual cleanup, there will be two libraries to load according to registry.
This is not fixable from driver or installation level.

(The other vendors do not have this issue only because they haven't switched to u-inf installation yet as far as I know.)

### DIFF
--- a/loader/icd.c
+++ b/loader/icd.c
@@ -79,12 +79,28 @@ void khrIcdVendorAdd(const char *libraryName)
         goto Done;
     }
 
+    // get the library's file name
+    const char *libName = libraryName;
+    const char *c;
+    for (c = libraryName; *c; ++c)
+    {
+        if (*c == DIRECTORY_SYMBOL)
+        {
+            libName = c + 1;
+        }
+    }
+
     // ensure that we haven't already loaded this vendor
     for (vendorIterator = khrIcdVendors; vendorIterator; vendorIterator = vendorIterator->next)
     {
         if (vendorIterator->library == library)
         {
             KHR_ICD_TRACE("already loaded vendor %s, nothing to do here\n", libraryName);
+            goto Done;
+        }
+        if (!strcmp(vendorIterator->libName, libName))
+        {
+            KHR_ICD_TRACE("already loaded library %s, nothing to do here\n", libName);
             goto Done;
         }
     }
@@ -184,6 +200,8 @@ void khrIcdVendorAdd(const char *libraryName)
             KHR_ICD_TRACE("failed get platform handle to library\n");
             continue;
         }
+        vendor->libName = (char *)malloc(strlen(libName) + 1);
+        strcpy(vendor->libName, libName);
         vendor->clGetExtensionFunctionAddress = p_clGetExtensionFunctionAddress;
         vendor->platform = platforms[i];
         vendor->suffix = suffix;
@@ -417,3 +435,15 @@ void khrIcdContextPropertiesGetPlatform(const cl_context_properties *properties,
     }
 }
 
+void khrIcdFreeLibName()
+{
+    KHRicdVendor *vendorIterator;
+    for (vendorIterator = khrIcdVendors; vendorIterator; vendorIterator = vendorIterator->next)
+    {
+        if (vendorIterator->libName != NULL)
+        {
+            free(vendorIterator->libName);
+            vendorIterator->libName = NULL;
+        }
+    }
+}
--- a/loader/linux/icd_linux.c
+++ b/loader/linux/icd_linux.c
@@ -145,6 +145,8 @@ void khrIcdOsVendorsEnumerate(void)
         closedir(dir);
     }
 
+    khrIcdFreeLibName();
+
     if (NULL != envPath)
     {
         khrIcd_free_getenv(envPath);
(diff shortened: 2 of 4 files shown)


loader/linux/icd_linux.c
```diff
@@ -143,6 +143,16 @@ void khrIcdOsVendorsEnumerate(void)
...
+
+        KHRicdVendor *vendorIterator;
+        for (vendorIterator = khrIcdVendors; vendorIterator; vendorIterator = vendorIterator->next)
+        {
+            if (vendorIterator->libName != NULL)
+            {
+                free(vendorIterator->libName);
+                vendorIterator->libName = NULL;
+            }
+        }
```
Kerilk, line 155: Code is duplicated here and in `icd_windows.c`. Also this is cleaning up something created in `icd.c` and is platform independent. This should go in it's own function defined in `icd.c` and called from here.

loader/windows/icd_windows.c
```diff
@@ -247,6 +247,17 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
...
+    KHRicdVendor *vendorIterator;
+    for (vendorIterator = khrIcdVendors; vendorIterator; vendorIterator = vendorIterator->next)
+    {
+        if (vendorIterator->libName != NULL)
+        {
+            free(vendorIterator->libName);
+            vendorIterator->libName = NULL;
+        }
+    }
+
```
Kerilk, line 260: See above.

Kerilk (changes requested):
This is a very impactful change, and I think it will need to be discussed. Especially, I don't know how many users could be loading several version of the same OpenCL implementation for legitimate uses.

Also I am not a Windows expert, so I can't really comment on the specifics. But if other vendors will encounter the same issue, they should be part of this discussion.

lttamd:
> This is a very impactful change, and I think it will need to be discussed. Especially, I don't know how many users could be loading several version of the same OpenCL implementation for legitimate uses.
> 
> Also I am not a Windows expert, so I can't really comment on the specifics. But if other vendors will encounter the same issue, they should be part of this discussion.

As far as I understand, it is wrong to load multiple versions of the same OpenCL implementation from the same vendor. (unlike loading OpenCL implementations from different vendors)
At least for AMD OpenCL, loading multiple amdocl binaries will definitely causes undefined behaviors. 

Also, even before this bug showing up as result of new u-inf installation using the new registry location, I don't think ICD ever allowed "multiple versions of the same OpenCL implementation" to be loaded at the same time. 
See the code from line 82 to line 90 in icd.c, ICD has been doing check to "ensure that we haven't already loaded this vendor"(quoted from the comment).
But it only checks library object pointer, so it won't really filter out all duplicates since amdocl now can come from different pointers(registry locations) because of the new u-inf installation.

Therefore, I do not think this is an impactful change, but a merely a bug fix following existing ICD loading rule. 
For AMD OpenCL, we have this change as workaround for many releases without causing any issue.
For other vendors, it should have no impact.

loader/icd.c
```diff
@@ -79,6 +79,17 @@ void khrIcdVendorAdd(const char *libraryName)
...
+    // get the library's file name
+    const char *libName = libraryName;
+    const char *c;
+    for (c = libraryName; *c; ++c)
+    {
+        if (*c == DIRECTORY_SYMBOL)
+        {
+            libName = c + 1;
+        }
+    }
```
bashbaug, line 91: This may be a non-issue, but so we go in with eyes open: As I understand this change strips all path information from the library name and only loads the first library with a given name.  While this will prevent unintentional scenarios where the same library name would be loaded multiple times, it also means that library names must be unique, even if they are in different directories, or the ICD loader will filter them out.  Is this the behavior we want?

I'm not aware of any implementors doing this, but a scenario I've been thinking of where two libraries could legitimately have the same name is if one device generation is supported by one software package and installed into one directory and another device generation is supported by a different package and installed into a different directory, but both device generations have the same library file name.

lttamd: Yes. This is the behavior we want. 
See my previous comment. ICD's rule on duplicate libraries has always been like this. 

There is no such scenario at least for us. (Support for multiple device generations is not implemented in this way.) 
Unlike loading OpenCL libraries form different vendors, loading multiple libraries from the same vendor is undefined behavior.

lttamd:
Do we have any update on this?

bashbaug:
As per my action item from the December 3rd teleconference, here were a few prior teleconferences where we discussed this issue:

* There was a very brief discussion on August 13th.
* There was an in-depth discussion the following week, on August 20th, but without any clear next steps or conclusions.
* I don't see any subsequent discussion in teleconferences or in the October f2f.

For next steps, perhaps we can converge on solution requirements?  Briefly, in prior discussions proposed requirements were:

1. Desire to order platforms – not just for single vendor
2. Exclusivity
3. Multiple platforms exposing the same device 

Is this correct?  If so, are some of these requirements more important than others and hence require a short-term solution, whereas others could be solved longer-term?

Issue #184, opened 2022-10-04
Improve FreeBSD Support


bashbaug:
see: https://github.com/KhronosGroup/OpenCL-SDK/issues/55

A minimum we should modify the OpenCL ICD loader so it checks for FreeBSD or Unix generically vs. just Linux, to fix build issues.

Separately we can investigate renaming some of the source files so they apply more generically to Unix vs. Linux.

juliohm:
Regarding the FreeBSD environment to test, the build script I provided in https://github.com/juliohm/cross-platform-opencl should be enough.

You can type

```
$ julia --project build_tarballs.jl x86_64-unknown-freebsd
```

to emulate a FreeBSD.

bashbaug:
Thanks!  I'll test your build script next.  I also setup a FreeBSD VM so I can at least build and run unit tests.

PR #185, opened 2022-10-04
add FreeBSD define to icd_platform.h

bashbaug:
Adds minimal support for compiling on FreeBSD.  See #184.

CC @rajhlinux @juliohm


Kerilk (approved):
Nice improvements here, thanks.

Issue #130, opened 2020-10-23
cross-compilation under Linux/Mingw64 fails after newer commits


hydra3333:
Hello, 

I am trying to cross-compile the ICL Loader under Linux/Mingw64 with a target of Win10x64 and Nvidia driver (openCL v1.2).

Previously, you very kindly created `-DOPENCL_ICD_LOADER_DISABLE_OPENCLON12=ON` so as to create an import lib,
https://github.com/KhronosGroup/OpenCL-ICD-Loader/commit/bb98ad9a9c264d63ad2204c1eeee0a3a1e724e20
"This variable should only be used when building an import lib to link with"

Then, recently, there have been ICD Loader commits which appear to have broken previous compatibility with that cross compilation.
I seemed to have fixed one of the issues by adding stuff at and near the "elseif(MINGW)" part of the patch below, so it can use `-DOPENCL_ICD_LOADER_DISABLE_OPENCLON12=ON`.
The other (slightly old) patches are copied from MingW packages at https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-opencl-icd-git

However, I still can't get it build any more.
The build errors are shown in the log below.

Any assistance would be greatly appreciated.

Thank you.

```
[DEBUG] patch -b -p1 < "01-mingw-build-Alexpux-updated-2020.10.23.patch"
patching file CMakeLists.txt

[DEBUG] patch -b -p1 < "01a-mingw-build-Alexpux-updated-2020.10.23.patch"
patching file Config.cmake.in

[DEBUG] patch -b -p1 < "02-mingw-build-Alexpux-updated-2020.10.23.patch"
patching file loader/windows/icd_windows_hkr.c

[DEBUG] patch -b -p1 < "03-mingw-build-Alexpux-updated-2020.10.23.patch"
patching file test/loader_test/test_create_calls.c

[DEBUG] patch -b -p1 < "04-mingw-build-Alexpux-updated-2020.10.23.patch"
patching file test/loader_test/test_program_objects.c

[...]

PR #186, opened 2022-10-07
OpenCL ICD Loader Fixes for mingw

### LINKED ISSUE #191: difficulties building with 32-bit mingw-i686
I've been trying to get the OpenCL ICD loader building and working properly with 32-bit mingw-i686 as part of https://github.com/KhronosGroup/OpenCL-SDK/issues/58 but I seem to have hit a bit of a dead end, so I'm at least documenting what I've tried in case it is helpful.

I'm cross-compiling on a Linux host (Ubuntu 20.04).  I have the mingw-w64 packages installed.

```sh
$ i686-w64-mingw32-g++ --version
i686-w64-mingw32-g++ (GCC) 9.3-posix 20200320
```

I'm trying to build my [SimpleOpenCLSamples](https://github.com/bashbaug/SimpleOpenCLSamples), which builds with the upstream OpenCL-Headers and OpenCL-ICD-Loader.  The specific commits I am using are:

* SimpleOpenCLSamples - https://github.com/bashbaug/SimpleOpenCLSamples/commit/42ca56455295414bb926f8aec9a2d32a8fe0af4f (latest in the main branch)
* OpenCL-Headers - https://github.com/KhronosGroup/OpenCL-Headers/commit/b590a6bfe034ea3a418b7b523e3490956bcb367a (latest in the main branch)
* OpenCL-ICD-Loader - https://github.com/bashbaug/OpenCL-ICD-Loader/commit/6c51a7912d2a57754f6e8bf1bb4159458768f2a1 (latest in a branch with a few mingw fixes/improvements)

With these commits I am able to build an OpenCL.dll and a libOpenCL.dll.a import library, but I get an error when I link even the simplest sample with these libraries:

```sh
$ make clean enumopencl
[  8%] Building C object external/opencl-icd-loader/CMakeFiles/OpenCL.dir/loader/icd.c.obj
...
[ 83%] Linking C shared library OpenCL.dll
[ 83%] Built target OpenCL
[ 91%] Building CXX object samples/00_enumopencl/CMakeFiles/enumopencl.dir/main.cpp.obj
[100%] Linking CXX executable enumopencl.exe
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0x148): undefined reference to `clGetPlatformInfo@20'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0x1ad): undefined reference to `clGetPlatformInfo@20'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0x350): undefined reference to `clGetDeviceInfo@20'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0x3b5): undefined reference to `clGetDeviceInfo@20'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0x5d2): undefined reference to `clGetDeviceInfo@20'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0x9af): undefined reference to `clGetPlatformIDs@12'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0xa17): undefined reference to `clGetPlatformIDs@12'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0xab2): undefined reference to `clGetDeviceIDs@24'
/usr/bin/i686-w64-mingw32-ld: CMakeFiles/enumopencl.dir/objects.a(main.cpp.obj):main.cpp:(.text+0xb3e): undefined reference to `clGetDeviceIDs@24'
collect2: error: ld returned 1 exit status
make[3]: *** [samples/00_enumopencl/CMakeFiles/enumopencl.dir/build.make:101: samples/00_enumopencl/enumopencl.exe] Error 1
make[2]: *** [CMakeFiles/Makefile2:1326: samples/00_enumopencl/CMakeFiles/enumopencl.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:1333: samples/00_enumopencl/CMakeFiles/enumopencl.dir/rule] Error 2
make: *** [Makefile:374: enumopencl] Error 2
```

The problem appears to be due to usage of the `stdcall` calling convention for 32-bit Windows, though I'm not 100% sure where things are going wrong.  Here is my current understanding:

1. The application itself is looking for a decorated function name (with the `@n` suffix for stdcall) and is not finding it, hence the linker failure.

2. The import library libOpenCL.dll.a does NOT decorate the functions it exports:

```sh
$ winedump -j export external/opencl-icd-loader/libOpenCL.dll.a 
Contents of external/opencl-icd-loader/libOpenCL.dll.a: 81250 bytes

248 public symbols
    1bd4 _OpenCL_dll_iname
    1d0e __head_OpenCL_dll
    1e8a _clWaitForEvents
    1e8a __imp__clWaitForEvents
    20d2 _clUnloadPlatformCompiler
    20d2 __imp__clUnloadPlatformCompiler
...
```

3. The DLL itself also does NOT decorate the functions it exports:

```sh
$ winedump -j export external/opencl-icd-loader/OpenCL.dll 
Contents of external/opencl-icd-loader/OpenCL.dll: 419689 bytes

Exports table:

  Name:            OpenCL.dll
  Characteristics: 00000000
  TimeDateStamp:   6344ADAB Mon Oct 10 16:41:31 2022
  Version:         0.00
  Ordinal base:    1
  # of functions:  123
  # of Names:      123
Addresses of functions: 00015028
Addresses of name ordinals: 00015400
Addresses of names: 00015214

  Entry Pt  Ordn  Name
  00003EE2     1 clBuildProgram
  000085AC     2 clCloneKernel
  00006ED5     3 clCompileProgram
...
```

(2) and (3) seem to be consistent with the OpenCL.def in this repo, which also does not decorate the functions it exports:

```
$ more ../external/opencl-icd-loader/loader/windows/OpenCL.def 
; copyright text removed here

EXPORTS

; comment removed here

; OpenCL 1.0 API
clBuildProgram
clCreateBuffer
clCreateCommandQueue
clCreateContext
...
```

I suppose this could still be the problem though?

Some things I tried:

* Adding the linker option `--add-stdcall-alias` (didn't seem to have an effect).
* Adding the linker option `--disable-stdcall-fixup` (couldn't link OpenCL.dll).
* Various combinations with the linker option `--kill-at` (didn't seem to have an affect).
* Changing OpenCL.def to explicitly add an alias to the decorated function (didn't seem to have an affect):

    ```
    clGetPlatformIDs = clGetPlatformIDs@12
    ```

* Changing OpenCL.def to only include the decorated function:

    ```
    clGetPlatformIDs@12
    ```

    I was able to link an executable with this change.  Not sure how viable of a solution this is, though...

bashbaug:
This PR has several fixes that enable building the OpenCL ICD loader and an import library using mingw:

1. For 32-bit mingw, using a special .def file with stdcall-decorated functions (e.g. `function@n`).  This is needed to generate a proper import library for 32-bit mingw only.  For all other builds we will continue to use the same .def file as before.
2. Change the HKR GUID query from `CM_Get_DevNode_PropertyW` to `CM_Get_DevNode_Registry_PropertyW`.  This performs the same query but using a function that is in mingw headers.  Note, this is the similar flow used by the Vulkan loader ([link]( https://github.com/KhronosGroup/Vulkan-Loader/blob/823078c862f2d6deff673ca8d45a591f620f5712/loader/loader_windows.c#L232)).
3. Updates `_WIN32_WINNT` to `0x0600` (Windows Vista) for `InitOnceExecuteOnce`.
4. Minor CMakeFile fix to build an `OpenCL.dll` instead of a `libOpenCL.dll`.

Note, I did need to disable OpenCLOn12 via `OPENCL_ICD_LOADER_DISABLE_OPENCLON12`, so this isn't quite perfect, but it's much better than what we have currently.

Fixes #191.

See also: #11? #130?

See also: https://github.com/KhronosGroup/OpenCL-SDK/issues/58

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,13 +80,20 @@ if (WIN32)
         loader/windows/icd_windows_hkr.h
         loader/windows/icd_windows_apppackage.c
         loader/windows/icd_windows_apppackage.h
-        loader/windows/OpenCL.def
         loader/windows/OpenCL.rc)
     # Only add the DXSDK include directory if the environment variable is
     # defined.  Since the DXSDK has merged into the Windows SDK, this is
     # only required in rare cases.
     if (DEFINED ENV{DXSDK_DIR} AND NOT (MINGW OR MSYS OR CYGWIN))
         include_directories ($ENV{DXSDK_DIR}/Include)
+    endif ()
+
+    # For mingw-i686 builds only we need a special .def file with stdcall
+    # exports.  In all other cases we can use a standard .def file.
+    if (("${CMAKE_SIZEOF_VOID_P}" EQUAL "4") AND (MINGW OR MSYS OR CYGWIN))
+        list (APPEND OPENCL_ICD_LOADER_SOURCES loader/windows/OpenCL-mingw-i686.def)
+    else ()
+        list (APPEND OPENCL_ICD_LOADER_SOURCES loader/windows/OpenCL.def)
     endif ()
 else ()
     list (APPEND OPENCL_ICD_LOADER_SOURCES
@@ -115,7 +122,7 @@ if (WIN32)
     # Generate a DLL without a "lib" prefix for mingw.
     if (MINGW OR MSYS OR CYGWIN)
         set_target_properties(OpenCL PROPERTIES PREFIX "")
-        set_target_properties(OpenCL PROPERTIES LINK_FLAGS "-Wl,-enable-stdcall-fixup")
+        set_target_properties(OpenCL PROPERTIES LINK_FLAGS "-Wl,-disable-stdcall-fixup")
     endif()
 else()
     target_link_libraries (OpenCL PRIVATE ${CMAKE_THREAD_LIBS_INIT})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,12 @@ if (WIN32)
     target_link_libraries (OpenCL PRIVATE cfgmgr32.lib runtimeobject.lib)
 
     option (OPENCL_ICD_LOADER_DISABLE_OPENCLON12 "Disable support for OpenCLOn12. Support for OpenCLOn12 should only be disabled when building an import lib to link with, and must be enabled when building an ICD loader for distribution!" OFF)
+
+    # Generate a DLL without a "lib" prefix for mingw.
+    if (MINGW OR MSYS OR CYGWIN)
+        set_target_properties(OpenCL PROPERTIES PREFIX "")
+        set_target_properties(OpenCL PROPERTIES LINK_FLAGS "-Wl,-enable-stdcall-fixup")
+    endif()
 else()
     target_link_libraries (OpenCL PRIVATE ${CMAKE_THREAD_LIBS_INIT})
     if (NOT APPLE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,7 @@ if (WIN32)
 
     # For mingw-i686 builds only we need a special .def file with stdcall
     # exports.  In all other cases we can use a standard .def file.
-    if (("${CMAKE_SIZEOF_VOID_P}" EQUAL "4") AND (MINGW OR MSYS OR CYGWIN))
+    if ((CMAKE_SIZEOF_VOID_P EQUAL 4) AND (MINGW OR MSYS OR CYGWIN))
         list (APPEND OPENCL_ICD_LOADER_SOURCES loader/windows/OpenCL-mingw-i686.def)
     else ()
         list (APPEND OPENCL_ICD_LOADER_SOURCES loader/windows/OpenCL.def)
--- a/loader/icd_platform.h
+++ b/loader/icd_platform.h
@@ -40,4 +40,11 @@
 #error Unknown OS!
 #endif
 
+#ifdef __MINGW32__
+#if !defined(_WIN32_WINNT) || (_WIN32_WINNT < 0x0600)
+#undef _WIN32_WINNT
+#define _WIN32_WINNT 0x0600
+#endif
+#endif // __MINGW32__
+
 #endif
--- a/loader/windows/OpenCL-mingw-i686.def
+++ b/loader/windows/OpenCL-mingw-i686.def
@@ -1,0 +1,162 @@
+;
+; Copyright (c) 2022 The Khronos Group Inc.
+;
+; Licensed under the Apache License, Version 2.0 (the "License");
+; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
+;     http://www.apache.org/licenses/LICENSE-2.0
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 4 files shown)


Kerilk:
Awesome work @bashbaug , thanks for cracking this one!

Small questions:
 - did you extract the decorations from a test application? Or compute it from the signature?
 - did you generate the def file?
   - if not should we? (To clarify, I am not saying it should be done in this PR)
   - if yes should we include the generator script? (Again, maybe not in this PR)

bashbaug:
> * did you extract the decorations from a test application? Or compute it from the signature?

I actually didn't need to do either of these, since the mingw linker printed a message when it couldn't find the undecorated name:

```
Warning: resolving _clEnqueueNDRangeKernel by linking to _clEnqueueNDRangeKernel@36
Warning: resolving _clEnqueueNativeKernel by linking to _clEnqueueNativeKernel@40
Warning: resolving _clEnqueueReadBuffer by linking to _clEnqueueReadBuffer@36
...
```

I've since cross-checked this against the exports created by MSVC and the output from the "gendef" tool so I'm fairly certain I got them all and they're correct, but it is still a bit of a manual process.  If we wanted to be more sure we could auto-generate a "call every API" test file and include it in CI, but that's getting a little ahead of ourselves since we don't even have this config in CI  (yet?).

> * did you generate the def file?

No, I looked into this but I didn't find any existing projects I could start from or obvious CMake functionality to help out.  I'd still like to come back to ths at some point.

The very best solution IMHO would be to use more `__declspec(dllexport)`-like functionality to avoid the need for .def files entirely, but I couldn't quite get this to do what I wanted to, either.

We could also look into a solution similar to [Mesa](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/bin/gen_vs_module_defs.py), where a single .def file gets processed by a python script, so at least then we don't need to keep two .def files in sync.  That's debatably overkill though, since we shouldn't need to modify our .def files very often.

Kerilk:
Thanks for the details @bashbaug . I'll take a shot at generating these files (def + map) using the python/Docs framework once this is merged. It seems the decoration is 4*number of arguments, and if there are exceptions they should be easy to spot with the diff.

Kerilk (approved):

Kerilk:
I approved, but I think @jenatali should take a look at the `loader/windows/icd_windows_hkr.c` modifications just to make sure I didn't overlook something.

jenatali (approved):
LGTM, though I don't know the CM APIs very well.

Kerilk:
@bashbaug do you want me to press the merge button, or do you think more discussion is needed?

bashbaug:
I think this is ready to go - I've been testing with these changes for some time now and I haven't seen any issues, and the only semi-risky part is the specific .def file, which is mingw specific.

PR #188, opened 2022-10-10
Implement clGetICDLoaderInfoOCLICD.

Kerilk:
This implements clGetICDLoaderInfoOCLICD as suggested in https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/182

With this change, `clinfo` reports:
```
ICD loader properties
  ICD loader Name                                 Khronos OpenCL ICD Loader
  ICD loader Vendor                               Khronos Group
  ICD loader Version                              3.0.4
  ICD loader Profile                              OpenCL 3.0
```

If we go ahead with this, the `cl_icdl.h` file should be moved to the header, and the `cl_icdl` extension should be properly defined.

~~I decided to keep the enumeration, rather than use defines as we usually do, because I am unsure of the side effects on big endian architectures. This will cause some headaches when defining the xml extension as I think no other enums are defined.~~


loader/icd_dispatch.c
```diff
@@ -18,9 +18,52 @@
...
+clGetICDLoaderInfoOCLICD(
+    cl_icdl_info param_name,
+    size_t       param_value_size,
+    void *       param_value,
+    size_t *     param_value_size_ret)
+{
+    static const char  cl_icdl_OCL_VERSION[] = CL_ICDL_OCL_VERSION_STRING;
+    static const char  cl_icdl_VERSION[] = OPENCL_ICD_LOADER_VERSION_STRING;
+    static const char  cl_icdl_NAME[] = CL_ICDL_NAME_STRING;
+    static const char  cl_icdl_VENDOR[] = CL_ICDL_VENDOR_STRING;
```
bashbaug, line 36: Do we need these local variables here?  If so, do they need to be "static"?

loader/icd_dispatch.c
```diff
@@ -18,9 +18,52 @@
...
+    size_t             pvs;
+    void *             pv;
+
+#define KHR_ICD_CASE_PARAM_NAME(name)                                          \
+    case CL_ICDL_ ## name:                                                     \
+        pvs = sizeof(cl_icdl_ ## name);                                        \
```
bashbaug, line 42: This will work as long as these are all string literals or arrays (which they are currently), but I'd still consider calling strlen instead to avoid possible problems in the future.

Or, maybe this is the reason for the local variables above?

loader/cl_icdl.h
```diff
@@ -0,0 +1,28 @@
...
+    CL_ICDL_VERSION     = 2,
+    CL_ICDL_NAME        = 3,
+    CL_ICDL_VENDOR      = 4,
+} cl_icdl_info;
+
+typedef cl_int
+(CL_API_CALL * pfn_clGetICDLoaderInfoOCLICD)(cl_icdl_info param_name,
+                                             size_t       param_value_size,
+                                             void *       param_value,
+                                             size_t *     param_value_size_ret);
```
bashbaug, line 21: I'm not sure if it's too late to change the names of these enums and functions in practice, but if it's not it would be nice to align them more closely to OpenCL naming conventions:

* For the enum names, could we be a bit more descriptive?  Maybe something like `CL_ICD_LOADER_OPENCL_VERSION` instead of `CL_ICDL_OCL_VERSION`?
    * Should these enums have a suffix like other extension enums, and if so, what suffix should we use?

* For the function name, is `OCLICD` part of the function name, or is this the extension suffix?  If it is the extension suffix then it should match whatever extension suffix we choose to use for the extension enums.  If it is part of the extension name then I think it is redundant, and the name `clGetICDLoaderInfo` is sufficient on its own.

* We should use a typedef for `cl_icdl_info` (or whatever we end up calling it) vs. an enum type.  We should also `#define` the values of each enum similarly.

CMakeLists.txt
```diff
@@ -137,6 +139,12 @@ endif ()
 target_compile_definitions (OpenCL
   PRIVATE
     CL_TARGET_OPENCL_VERSION=300
+    CL_ICDL_OCL_VERSION_STRING="OpenCL 3.0"
```
bashbaug, line 142: Does it make sense to build this string similar to the OPENCL_ICD_LOADER_VERSION_STRING as "OpenCL " + major version + "." minor version vs. passing it explicitly?  This could be important if we drive more versioning from the CMake project version (#182).

bashbaug (commented):
Thanks for pulling this together - there are a lot of interesting ideas here!

loader/cl_icdl.h
```diff
@@ -0,0 +1,28 @@
+#ifndef OPENCL_CL_ICDL_H
+#define OPENCL_CL_ICDL_H
```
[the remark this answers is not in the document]
Kerilk: Indeed, but for CI purposes it always ends up being a 2 step process.

CMakeLists.txt
```diff
@@ -137,6 +139,12 @@ endif ()
...
+    OPENCL_ICD_LOADER_VERSION_MAJOR=3
+    OPENCL_ICD_LOADER_VERSION_MINOR=0
+    OPENCL_ICD_LOADER_VERSION_REV=4
+    CL_ICDL_NAME_STRING="Khronos OpenCL ICD Loader"
+    CL_ICDL_VENDOR_STRING="Khronos Group"
```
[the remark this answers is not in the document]
Kerilk: This was to map the enum values and make clear which corresponds to which.
Not sure this is a compelling argument to keep them. I'll update the names.

CMakeLists.txt
Kerilk: Yes, I see your point. I'll look into it.

loader/icd_dispatch.c
Kerilk: As said above, I don't feel strongly about the implementation. I didn't have a particular template to follow, so I studied a few implementations (including ocl-icd) and came up with this, which as you justly remarked works if data are stored in array like variables (integers would, for instance, need to be defined as `static const int a[1] = {CL_ICDL_XXX};`).

The only opinion I have against `strlen` based approach is that they cannot be generalized to other parameter types, and it means adding new options will add type specific variants of the `KHR_ICD_CASE_PARAM_NAME` macro.

loader/cl_icdl.h
Kerilk: As an addendum, for the type of the enum and why I didn't want to change it: the size of the enum is not guaranteed, and there may be platforms where changing to a cl_uint may break ABI. (I know this is supposed to be a theoretical problem... but still)

CMakeLists.txt
Kerilk: I pushed https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/188/commits/01c9c3136ec26a10630ecca4da2ffadde1a42ac4 that partially address this comment. `OPENCL_ICD_LOADER_OCL_VERSION_STRING` is really a reflect of the `CL_TARGET_OPENCL_VERSION` used to compile the loader and the API exported. The issue is OpenCL does not define MINOR/MAJOR macros that we could use to get the string from `CL_TARGET_OPENCL_VERSION`. We could experiment in the `loader/icd_version.h` file and maybe provide those in the official headers once we are satisfied.

Kerilk:
GitHub action `latest` images seem to have been updated, and CI is breaking.

loader/cl_icdl.h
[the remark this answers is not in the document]
bashbaug: Understood, I'd like to ensure we have a roadmap to get this into the OpenCL headers eventually though, assuming that is our goal.

What if we did something like this:

1. Rename this file for now, so it is clear it is a private or internal ICD loader extension, e.g. `private_cl_icdl.h`.

Then, assuming we want to add this to the OpenCL headers:

2. Add `cl_icdl.h` (or whatever we end up call it) with the same contents and merge to OpenCL-Headers.
3. After merging, remove the private file from this repo, and switch to the file from OpenCL-Headers instead.

Would that work?  I'm totally open to other suggestions too.

loader/icd_dispatch.c
bashbaug: Ah, I was originally thinking this would keep around some extra variables after the function returned but it doesn't look like that's the case, sorry for the distraction.

loader/icd_dispatch.c
bashbaug: The case I'd be a little worried about is the `foo3` case here: https://godbolt.org/z/oaEK6jbxr

This case returns size == 8 (the size of the pointer) rather than the length of the string.

Perhaps this isn't worth worrying about though...

loader/cl_icdl.h
bashbaug: Yeah, not a problem, I definitely see the value of working with shipping software so I'm personally fine with this as-is.

I do hope we can align more closely to OpenCL naming conventions if we choose to add similar ICD loader extensions in the future.  In other words, let's think of this as a one-off solution rather than establishing precedent.

I do think it would be wise to change from an enum to a fixed-size type (probably `cl_uint`) though.  As you mentioned, the size of an enum is not guaranteed, though this is determined by the compiler and not the platform - eek.

https://stackoverflow.com/questions/366017/what-is-the-size-of-an-enum-in-c

CMakeLists.txt
Kerilk: Yes you could build a recent loader ( layers, trace, etc) but target OpenCL 2.0. It's version would still be 3.0.4, but the OCL_VERSION would be 2.0

Kerilk: To expand a bit this is the reason for the different VERSION constants:
```C
#define CL_ICDL_OCL_VERSION 1
#define CL_ICDL_VERSION     2
```
The first is the OpenCL version supported, the second the version of the loader. For ocl-icd, the first returns "OpenCL 3.0" while the second returns "2.3.1"

loader/icd_dispatch.c
Kerilk: I update the macro, the assembly is the same. Tell me if you prefer it this way.

Kerilk:
@MathiasMagnus any idea why the MacOS CI fails here:
https://github.com/KhronosGroup/OpenCL-ICD-Loader/actions/runs/3439072786
but not there:
https://github.com/Kerilk/OpenCL-ICD-Loader/actions/runs/3439072644

loader/icd_dispatch.c
bashbaug: Thanks!  I'm honestly fine either way.  I like the new macro name `KHR_ICD_CASE_STRING_PARAM_NAME` now though, and I think this will help avoid potential problems in the future.  Let's go with this for now and we can always revisit if needed.

CMakeLists.txt
bashbaug: Makes sense, thanks for the explanation!

bashbaug (approved):
Thanks, all of my comments are resolved.

I think this needs a rebase now before merging.

Should we bump the rev number to 3.0.5?

Kerilk:
@bashbaug I have rebased the code on main, and bumped version number to 3.0.5. Unless we want to do the header integration first I think it is good to go.

PR #189, opened 2022-10-10
clean up usage of CL_API_ENTRY

bashbaug:
While experimenting with different mechanisms of exporting symbols from a DLL I found some discrepancies how we are using CL_API_ENTRY in the OpenCL ICD loader.  These aren't causing any problems right now because by default CL_API_ENTRY is defined to be empty and we are not changing the default, but it would be nice to clean these up in case we ever do decide to use CL_API_ENTRY.

### DIFF
--- a/loader/icd.h
+++ b/loader/icd.h
@@ -33,6 +33,18 @@
 #define CL_USE_DEPRECATED_OPENCL_1_2_APIS
 #endif
 
+#ifndef CL_USE_DEPRECATED_OPENCL_2_0_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_0_APIS
+#endif
+
+#ifndef CL_USE_DEPRECATED_OPENCL_2_1_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_1_APIS
+#endif
+
+#ifndef CL_USE_DEPRECATED_OPENCL_2_2_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_2_APIS
+#endif
+
 #include <CL/cl.h>
 #include <CL/cl_ext.h>
 #include <CL/cl_icd.h>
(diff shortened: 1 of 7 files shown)


loader/icd.h
```diff
@@ -33,6 +33,18 @@
...
+#endif
+
+#ifndef CL_USE_DEPRECATED_OPENCL_2_1_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_1_APIS
+#endif
+
+#ifndef CL_USE_DEPRECATED_OPENCL_2_2_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_2_APIS
+#endif
+
```
bashbaug, line 47: Note: We're currently including the OpenCL headers in two places in this project: in icd.h here and in icd_dispatch.h.  As a further improvement it would be nice to reduce this duplication.  Things we could consider are:

1. Have icd_dispatch.h include icd.h to pick up the headers.  This is basically already done in practice, because we're always including icd.h and icd_dispatch.h as a pair.
2. Move the dispatch table structures into icd.h and remove icd_dispatch.h.
3. Extract the OpenCL header includes into a third file and include it from both icd.h and icd_dispatch.h.

Kerilk:
@bashbaug I assume the `CL_API_ENTRY` is made to be used to add `__declspec( dllexport )` and similar. As such is doesn't make sense on functions not exported as part of the OpenCL library.

bashbaug:
> I assume the CL_API_ENTRY exists to add __declspec( dllexport ) and similar. As such is doesn't make sense on functions not exported as part of the OpenCL library.

Yes, I think that is the intent, and these were the experiments I was trying.

Thanks for fixing the layer generation script!

Kerilk (approved):

Kerilk:
I assume getting a green here would need rebasing those changes on main. I checked the sources and cannot find another instance of improper `CL_API_ENTRY` usage. So I think it is good to go.

@bashbaug please, tell me if you think it is ready to merge, and I will do so.

loader/icd.h
Kerilk: @bashbaug Do you want to address this before we merge, or should I go ahead?

bashbaug:
> please, tell me if you think it is ready to merge, and I will do so.

I think this is all set - I merged in the latest CI fixes just to be sure and everything is clean now.  Thanks!

PR #193, opened 2022-10-14
fix an incorrect KHR_ICD_ASSERT

### LINKED ISSUE #192: failed assert: len == (sizeof(cszLibraryName) - 1)
Found as part of the mingw enabling.  This OpenCL ICD loader assert is currently failing:

```c++
char cszLibraryName[MAX_PATH] = { 0 };
...
size_t len;
wcstombs_s(&len, cszLibraryName, sizeof(cszLibraryName), pWchar, sizeof(cszLibraryName));
KHR_ICD_ASSERT(len == (sizeof(cszLibraryName) - 1));
```

I think this is a problem with the assert, not a problem with the code:

* After the call to [wcstombs_s](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/wcstombs-s-wcstombs-s-l?view=msvc-170) `len` should contain "The size in bytes of the converted string, including the null terminator."
* But, `sizeof(cszLibraryName)` is the size of the array irrespective of the string it stores, so it is always 260 (the value of `MAX_PATH`).

Possible alternative asserts:

1. KHR_ICD_ASSERT(len != 0): a return value of zero indicates that the conversion was unsuccessful.
2. KHR_ICD_ASSERT(len == strlen(cszLibraryName) + 1): though the value of this assert is questionable?
3. KHR_ICD_ASSERT(len == wcslen(pWchar)): might require an additional include?

Other notes:

* We should change the KHR_ICD_ASSERT macro so it adds a newline to the trace message (currently missing).
* We have an off-by-one error: the "count" should not include the terminating NULL character, so the call to wctombs_s should be:

```c++
wcstombs_s(&len, cszLibraryName, sizeof(cszLibraryName), pWchar, sizeof(cszLibraryName) - 1);
```

bashbaug:
fixes #192 

* Adds a newline to the assert trace message.
* Passes the correct "count" to [wcstombs_s](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/wcstombs-s-wcstombs-s-l?view=msvc-170) - should not include the terminating NULL character.
* Changes the actual assert itself to something meaningful.

### DIFF
--- a/loader/windows/icd_windows_dxgk.c
+++ b/loader/windows/icd_windows_dxgk.c
@@ -44,7 +44,6 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
         LoaderEnumAdapters2 EnumAdapters;
         NTSTATUS status = STATUS_SUCCESS;
 
-        char cszLibraryName[MAX_PATH] = { 0 };
         EnumAdapters.adapter_count = 0;
         EnumAdapters.adapters = NULL;
         PFN_LoaderEnumAdapters2 pEnumAdapters2 = (PFN_LoaderEnumAdapters2)GetProcAddress(h, "D3DKMTEnumAdapters2");
@@ -146,12 +145,22 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
             }
             if (NT_SUCCESS(status) && pQueryArgs->status == LOADER_QUERY_REGISTRY_STATUS_SUCCESS)
             {
-                wchar_t* pWchar = pQueryArgs->output_string;
-                memset(cszLibraryName, 0, sizeof(cszLibraryName));
+                char cszLibraryName[MAX_PATH];
+                result = WideCharToMultiByte(
+                    CP_ACP,
+                    0,
+                    pQueryArgs->output_string,
+                    -1,
+                    cszLibraryName,
+                    MAX_PATH,
+                    NULL,
+                    NULL);
+                if (!result)
                 {
-                    size_t len;
-                    wcstombs_s(&len, cszLibraryName, sizeof(cszLibraryName), pWchar, sizeof(cszLibraryName) - 1);
-                    KHR_ICD_ASSERT(len == wcslen(pWchar) + 1);
+                    KHR_ICD_TRACE("WideCharToMultiByte status != SUCCESS\n");
+                }
+                else
+                {
                     ret |= adapterAdd(cszLibraryName, EnumAdapters.adapters[AdapterIndex].luid);
                 }
             }
--- a/loader/windows/icd_windows_dxgk.c
+++ b/loader/windows/icd_windows_dxgk.c
@@ -98,7 +98,7 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
             queryArgs.query_flags.translate_path = TRUE;
             queryArgs.value_type = REG_SZ;
             result = MultiByteToWideChar(
-                CP_ACP,
+                CP_UTF8,
                 0,
                 cszOpenCLRegKeyName,
                 szOpenCLRegKeyName,
@@ -147,7 +147,7 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
             {
                 char cszLibraryName[MAX_PATH];
                 result = WideCharToMultiByte(
-                    CP_ACP,
+                    CP_UTF8,
                     0,
                     pQueryArgs->output_string,
                     -1,
--- a/loader/windows/icd_windows_dxgk.c
+++ b/loader/windows/icd_windows_dxgk.c
@@ -150,8 +150,8 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
                 memset(cszLibraryName, 0, sizeof(cszLibraryName));
                 {
                     size_t len;
-                    wcstombs_s(&len, cszLibraryName, sizeof(cszLibraryName), pWchar, sizeof(cszLibraryName));
-                    KHR_ICD_ASSERT(len == (sizeof(cszLibraryName) - 1));
+                    wcstombs_s(&len, cszLibraryName, sizeof(cszLibraryName), pWchar, sizeof(cszLibraryName) - 1);
+                    KHR_ICD_ASSERT(len == wcslen(pWchar) + 1);
                     ret |= adapterAdd(cszLibraryName, EnumAdapters.adapters[AdapterIndex].luid);
                 }
             }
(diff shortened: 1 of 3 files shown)


loader/windows/icd_windows_dxgk.c
```diff
@@ -150,8 +150,8 @@ bool khrIcdOsVendorsEnumerateDXGK(void)
                 memset(cszLibraryName, 0, sizeof(cszLibraryName));
                 {
                     size_t len;
-                    wcstombs_s(&len, cszLibraryName, sizeof(cszLibraryName), pWchar, sizeof(cszLibraryName));
-                    KHR_ICD_ASSERT(len == (sizeof(cszLibraryName) - 1));
+                    wcstombs_s(&len, cszLibraryName, sizeof(cszLibraryName), pWchar, sizeof(cszLibraryName) - 1);
+                    KHR_ICD_ASSERT(len == wcslen(pWchar) + 1);
```
Kerilk, line 154: @bashbaug please bear with me here, I am not a Windows expert, but I have read the documentation and have a nagging doubt. I am unsure this can work since (emphasis mine):

> **`wcstombs_s(size_t *pReturnValue, char *mbstr, size_t sizeInBytes, const wchar_t *wcstr, size_t count);`**
> `pReturnValue`
> The size in **bytes** of the converted string, including the null terminator.

> **`wcslen`**
> 1) Returns the length of a wide string, that is the **number** of non-null wide characters that precede the terminating null wide character.

So we're comparing bytes (as string which may include multi-bytes characters) to string length on wide characters. So if we have a wide character (something that results in a multibytes character) in the mix shouldn't that make `len` bigger and the assert fail, even if conversion was correct and successful?

Am I missing something here?

bashbaug: Yeah, full disclosure: I'm not 100% sure if this is the right assert either, but here's how the thinking goes (and the assert does pass, at least in my limited testing):

If each character in the wide-character string pWchar gets converted to a single-byte character in the multi-byte string cszLibraryName then the assert is fine and it will pass.  This is probably the common case... but I suppose it isn't guaranteed either, hrm.

If we're not comfortable with this assumption I suppose we could call [_mbslen](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strlen-wcslen-mbslen-mbslen-l-mbstrlen-mbstrlen-l?view=msvc-170) instead , which will return the number of characters in the multi-byte string, so (`KHR_ICD_ASSERT(_mbslen(cszLibraryName) == wcslen(pWchar))`).  Although, I am a little nervous about the "important note" in the docs:

> _mbslen, _mbslen_l, _mbstrlen, and _mbstrlen_l cannot be used in applications that execute in the Windows Runtime.

I'm a bit at a loss because I'm trying to infer what this assert is trying to test for.  Should we be testing for a successful return value from wctombs_s instead (non-zero)?  Is any assert required at all?

Kerilk: I agree that testing success is the only thing we should do.

bashbaug: Please take a look a the latest changes:

* I changed from a call to wctombs_s to a call to WideCharToMultiByte.  This matches behavior elsewhere in the OpenCL ICD loader and is consistent with the call to MultiByteToWideChar earlier in this function.
* I changed from a KHR_ICD_ASSERT to a check for success and a KHR_ICD_TRACE if the conversion failed, which is also consistent with the rest of the OpenCL ICD loader.

Kerilk: Changes look great!

I am not a security expert, so I am not sure about this paragraph from the documentation (emphasis mine):

`[in] dwFlags`
`WC_NO_BEST_FIT_CHARS`
> Translate any Unicode characters that do not translate directly to multibyte equivalents to the default character specified by lpDefaultChar.  In other words, if translating from Unicode to multibyte and back to  Unicode again does not yield the same Unicode character, the function  uses the default character. This flag can be used by itself or in  combination with the other defined flags. For **strings that require validation, such as file**, resource, and user  names, the application should always use the WC_NO_BEST_FIT_CHARS flag.  This flag prevents the function from mapping characters to characters  that appear similar but have very different semantics. In some cases,  the semantic change can be extreme. For example, the symbol for "∞"  (infinity) maps to 8 (eight) in some code pages.

No idea if this is relevant in our context.

Kerilk:
As a side note, `KHR_ICD_ASSERT` is now unused, so maybe we should retire it. I find it's name misleading as it does not `assert` in practice and just prints a warning (that is only visible when tracing is active).

bashbaug:
> As a side note, `KHR_ICD_ASSERT` is now unused, so maybe we should retire it.

Good call, done.  We can always bring it back if we find a use for it...

loader/windows/icd_windows_dxgk.c
bashbaug: > No idea if this is relevant in our context.

Yeah, I'm not sure either.  I did a quick grep through the OpenCL ICD loader and the Vulkan loader and they both are currently passing zero for `dwFlags`, so they do not include `WC_NO_BEST_FIT_CHARS`.

@jenatali do you have any experience in this area?

One delta I did notice between the OpenCL ICD loader and the Vulkan loader is that we are currently passing `CP_ACP` as our `CodePage` whereas Vulkan is consistently passing `CP_UTF8`, so this would be good to confirm also.

jenatali: I've got pretty minimal unicode experience, personally, and I've never used `WC_NO_BEST_FIT_CHARS` before.

As for `CP_ACP` vs `CP_UTF8`, you probably want `UTF8` unless there's a reason you want ASCII only.

bashbaug: Thanks, I've switched from `CP_ACP` to `CP_UTF8` and my minimal testing still seems to be passing.

Kerilk (approved):

loader/windows/icd_windows_dxgk.c
Kerilk: @jenatali can we consider the issue resolved? I think with your approval we can merge.

jenatali (approved):

Kerilk:
@jenatali Thanks!
@bashbaug I think we're good on this one as well, tell me if you want me to merge.

bashbaug:
> tell me if you want me to merge.

That'd be great, no issue on my end.  Thanks!

PR #195, opened 2022-10-19
Cache Headers for CI

Kerilk:
This PR caches the result of the Headers install rather than checking them out, configuring them, and rebuilding them (and their tests) for each job. The cache depends on the main last commit, and thus should be refreshed if the main branch of the Headers is updated.

This PR yields significant CI performance improvements.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -3,7 +3,71 @@ name: Linux
 on: [push, pull_request]
 
 jobs:
+  headers:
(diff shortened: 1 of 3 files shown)


bashbaug:
Out of curiosity, did you do any experiments disabling testing for the headers (by passing `-DBUILD_TESTING=0` to CMake when building the headers)?  I'm wondering if that will provide similar benefits without requiring so many changes to CI.

PR #196, opened 2022-10-25
add extension API function pointer checks

bashbaug:
This PR adds checks to ensure that an extension function pointer in the ICD dispatch table is non-NULL before calling it.  If a function pointer is NULL then CL_INVALID_VALUE is returned.  This improves robustness for applications that assume they can safely call any non-NULL function pointer returned by clGetExtensionFunctionAddress or clGetExtensionFunctionAddressForPlatform.

This is a fix for one of the extension-related issues I described in the Khronos face-to-face meetings last week.


Kerilk:
Thanks for these changes @bashbaug they look great. They raise a small question:
 * I know we use the `CL_INVALID_OPERATION` in this way in all our projects (at least I do as well), but:
   - Has this error code been described in this way formally?
   - Do we want to introduce a new one?
   - Should we add it to the functions documentation?
   - If we keep the same error code, are there any ambiguity with functions that would be returning the same error code?

As a side note, and sorry to derail the conversation a bit, if you think the indirect dispatch proposal I made here:
https://github.com/Kerilk/OpenCL-ICD-Loader/compare/8031bf6aa3cf3e59c4597983c0410a7bbde725ec...a4f825c46669c0d15365dba1018bebed74d8ac4a
is sensible (without updating each object - just the platform dispatch table -, given that vendors are unlikely to copy the dispatch table into each new object, so each object should inherit from the modified clUnloadCompiler pointer), then, in this scenario, we could use loader provided stub functions that return the error.

bashbaug:
Thanks @kerilk for the insightful comments, as always.

I wanted these changes to be unambiguously "safe" which meant that I made a couple of design tradeoffs and therefore they are a bit more "conservative" than they might be otherwise:

1. I used an existing error code (`CL_INVALID_OPERATION`) vs. defining some other error code.  Note however that the spec description for [`CL_INVALID_OPERATION`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_OPERATION) is: "This is a generic error code that is returned when the requested operation is not a valid operation." so it doesn't seem _too_ inappropriate.
2. I added the additional runtime check per extension API call vs. installing a stub function or otherwise mucking around with the implementation-provided dispatch table.

For (1), I'm absolutely willing to switch to a more descriptive error code if we choose to add one (`CL_UNSUPPORTED` or similar).  We'd need to decide if the extra descriptiveness is worth adding a new error code.  There aren't a lot of extension APIs that currently return `CL_INVALID_OPERATION` and for those that do I think it'd be fairly easy to disambiguate why the error occurred, but of course there is no guaratnee this will hold into the future.

For (2), I get a little nervous playing tricks with implementation-provided dispatch tables (like using the dispatch table entry for `clUnloadCompiler` for an ICD loader dispatch table) without formally documenting what the ICD loader is and is not allowed to do.  Maybe I'm too paranoid, but I can think of a bunch of different ways where this could conceivably break and it wouldn't be clear that an ICD is doing anything wrong.

It's worth mentioning that the tricks with the ICD loader dispatch table does improve robustness vs. the `NULL` pointer checks in this PR in several cases, so this is definitely a problem worth investigating in more detail:

* An ICD for a very early version of OpenCL (say, OpenCL 1.0) may only provide a short `cl_icd_dispatch` structure, in which case the ICD loader check for a non-`NULL` extension API function pointer may "pass" if whatever is located beyond the short dispatch table happens to be non-`NULL`, or may crash if there is no mapped memory beyond the short dispatch table.
* The changes in this PR only test for extension APIs but the same problems could also occur for core APIs.

Kerilk (approved):
This is definitely an improvement, and we are aware of the tradeoffs, so looks good to me.

bashbaug:
This has been approved for a while now.  Merging.

PR #198, opened 2022-11-10
fix MacOS CI

bashbaug:
The GitHub actions macos-latest workflow was updated to macos-12 recently, which only has gcc-11.  Switch back to macos-11 for now at least, to unblock MacOS CI.

Also updates the checkout action to v3 to eliminate CI warnings.

edit: macos-12 only has gcc-11, not gcc-12.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -23,10 +23,10 @@ jobs:
 
     steps:
     - name: Checkout OpenCL-ICD-Loader
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: Checkout OpenCL-Headers
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         repository: KhronosGroup/OpenCL-Headers
         path: external/OpenCL-Headers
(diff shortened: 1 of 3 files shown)


Kerilk:
Thanks for fixing it! I supposed it was the issue, but GitHub action documentation is outdated and still lists macos-11 as latest. Apparently it is also not updated for everyone, as I am still getting macos-11 for my fork. I'll approve once I see a working test on MacOS.

Kerilk:
Got to love github actions: "Error: Unable to resolve action `actions/checkout@3`, unable to find version `3`" but only for MacOS
Maybe rollback to `actions/checkout@2` for MacOS only?

bashbaug:
Yeah, I'm pretty sure this is it.  I'm not sure exactly when the switch happened and it still looks like the docs are catching up, but there is some information and the announcement here:

https://github.com/actions/runner-images/issues/6384

I had a typo in my change to update the checkout actions (now fixed), but you can see the previous commit was successful.  I'm pretty confident this latest update should be clean.

Kerilk:
Looks good, I'll let it run approve and merge, this will unblock the rest.

Kerilk (approved):

PR #206, opened 2023-02-07
Fix VERSIONINFO for llvm-rc

### LINKED ISSUE #155: Can't compile OpenCL.rc on Windows using llvm-rc
llvm-rc (which is LLVM's replacement for CL's internal RC compiler) throws an error when compiling `loader/windows/OpenCL.rc`:
```
llvm-rc: Error in VERSIONINFO statement (ID 1):
Non-ASCII 8-bit codepoint (169) can't occur in a non-Unicode string
```
This refers to line 46:
```c
VALUE "LegalCopyright", "Copyright \251 The Khronos Group Inc 2016-2020"
```
llvm-rc seems not to understand `\251` (CL compiles this one just fine), though replacing it with a `(c)` fixes it:
```c
VALUE "LegalCopyright", "Copyright (c) The Khronos Group Inc 2016-2020"
```

To reproduce:
- Install ninja-build and LLVM (version 12.0.1, which is the latest stable version as of now)
- Run `cmake -B build -G Ninja` (make sure clang is in PATH)
- Run `cmake --build build` and see the error

<br>
<br>
<br>

Small sidenote:
clang warns (`-Wincompatible-pointer-types`) about line 196 in `loader/windows/icd_windows.c`:
```c
HRESULT hr = pCreateDXGIFactory(&IID_IDXGIFactory, &pFactory);
```
The problem here is `&pFactory` which (including `&`) is an `IDXGIFactory**`, though `CreateDXGIFactory` expects a `void**`.
This could be fixed by casting `pFactory` to `void**`:
```c
HRESULT hr = pCreateDXGIFactory(&IID_IDXGIFactory, (void**)&pFactory);
```
While this warning isn't really of importance (since it doesn't affect anything), it's quite easy to fix.

yeah-its-gloria:
Closes #155. Also fixes a printf format warning.

### DIFF
--- a/loader/windows/icd_windows.c
+++ b/loader/windows/icd_windows.c
@@ -423,7 +423,7 @@ void *khrIcdOsLibraryLoad(const char *libraryName)
     }
     if (!hTemp)
     {
-        KHR_ICD_TRACE("Failed to load driver. Windows error code is %d.\n", GetLastError());
+        KHR_ICD_TRACE("Failed to load driver. Windows error code is %lu.\n", GetLastError());
     }
     return (void*)hTemp;
 }
--- a/loader/windows/icd_windows.c
+++ b/loader/windows/icd_windows.c
@@ -423,7 +423,7 @@ void *khrIcdOsLibraryLoad(const char *libraryName)
     }
     if (!hTemp)
     {
-        KHR_ICD_TRACE("Failed to load driver. Windows error code is %lu.\n", GetLastError());
+        KHR_ICD_TRACE("Failed to load driver. Windows error code is %"PRIuDW".\n", GetLastError());
     }
     return (void*)hTemp;
 }
--- a/loader/windows/OpenCL.rc
+++ b/loader/windows/OpenCL.rc
@@ -33,7 +33,7 @@ BEGIN
         BEGIN
             VALUE "FileDescription" ,"OpenCL Client DLL"
             VALUE "ProductName"     ,OPENCL_ICD_LOADER_NAME_STRING
-            VALUE "LegalCopyright"  ,"Copyright \251 The Khronos Group Inc 2016-2020"
+            VALUE "LegalCopyright"  ,L"Copyright \251 The Khronos Group Inc 2016-2020"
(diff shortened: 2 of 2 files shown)


loader/windows/icd_windows.c
```diff
@@ -423,7 +423,7 @@ void *khrIcdOsLibraryLoad(const char *libraryName)
     }
     if (!hTemp)
     {
-        KHR_ICD_TRACE("Failed to load driver. Windows error code is %d.\n", GetLastError());
+        KHR_ICD_TRACE("Failed to load driver. Windows error code is %lu.\n", GetLastError());
```
bashbaug, line 426: I'm OK with this as-is and this is definitely an improvement, but a slightly more portable option would be to use `PRIuDW` here instead.  We used this a couple of other places in the OpenCL ICD loader to print `DWORD` values. 
 Example:
```
KHR_ICD_TRACE("Reading value %"PRIuDW"...\n", dwIndex);
```

bashbaug (approved):
Thank you for your contribution!  I have one minor comment - not a must-fix but a nice-to-have.

loader/windows/icd_windows.c
yeah-its-gloria: Fixed this. 👍

Kerilk:
Thanks for this contribution! I have a small question:

Disclaimer: I am no expert in Windows programming, so I may be saying something irrelevant here.

When looking at:
https://learn.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block
It seems that `"040904E4"` means that the values should be in `U.S. English` and in `Multilingual` character set (0x04E4 == 1252).
Thus shouldn't all the strings be prefixed by L here? Are we mixing character sets? Tagging @jenatali .

yeah-its-gloria:
By default, all strings are treated as ASCII in llvm-rc, and MSVC's RC automatically turns strings that use codepoints beyond that set into UTF-16, which llvm-rc doesn't.

bashbaug:
I filed https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/209 to track the question about string prefixes.  I'm going to merge this now though, since this is definitely an improvement over what we have.  Thank you for your contribution!

PR #214, opened 2023-04-01
fix Linux and MacOS CI

bashbaug:
This PR adds several CI-related fixes and improvements for MacOS and Linux (and one small Windows improvement):

* Removes the Ubuntu 18.04 configs.  This will stop working soon (tomorrow?) anyhow, as the Ubuntu 18.04 runner is deprecated.
* Explicitly installs gcc on MacOS if it isn't already installed.  Needed because gcc-9 is no longer installed by default.
* Don't bother building testing for OpenCL-Headers to speed up builds.

I moved the Debug and Release configs to Ubuntu 20.04 so we will have the same coverage.

I didn't move the Unix Makefile configs to Ubuntu 20.04 because it doubles the number of configs and I don't think it buys us much.  We may want to drop support for multiple generators from MacOS and Windows CI for the same reason.

If we're going to add anything more I think it should be testing for older versions of CMake, since now we're only testing CMake 3.21.2.


Kerilk (approved):
We need a working CI, so I think this is good. For the moment, I am torn between merging right away, and waiting for @MathiasMagnus review.

Kerilk:
@bashbaug would you want me to try and make the same proposal then here: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/216 but for this repository?

bashbaug:
> would you want me to try and make the same proposal

Sure, if you'd like, that'd be great!  I think expanding the CMake coverage is most important, as I mentioned in the PR description.

Would you be OK doing this as a separate PR?  It would be nice to merge this ASAP to fix CI, since it's holding up some other changes.

Kerilk:
Yeah, exactly my issue. But if we don't do it right away, it is not as urgent, and Maybe @MathiasMagnus would do it better. My point was trying avoiding tearing things down and rebuilding afterwards, which makes the review process difficult.
If we take this route, I can press the button here.

Tell me how you want to proceed.

bashbaug:
Closing in favor of https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/215

PR #216, opened 2023-04-10
replaces calls to LoadLibrary with explicit calls to LoadLibraryA

### LINKED ISSUE #208: Incorrect usage of LoadLibrary and string literal
Must be `LoadLibraryA("dll_name")` or `LoadLibrary(_T("dll_name"))`.

https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/2cde5d09953a041786d1cfdcb1c08704a82cb904/loader/windows/icd_windows.c#L276
https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/2cde5d09953a041786d1cfdcb1c08704a82cb904/loader/windows/icd_windows_dxgk.c#L38
https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/2cde5d09953a041786d1cfdcb1c08704a82cb904/loader/windows/icd_windows_apppackage.c#L42

bashbaug:
fixes #208 

Since we're unconditionally passing an ascii string, unconditionally call LoadLibraryA instead of LoadLibrary.  This ensures that the correct function is called even if somebody compiles the OpenCL ICD loader with UNICODE support.

### DIFF
--- a/loader/windows/icd_windows.c
+++ b/loader/windows/icd_windows.c
@@ -273,7 +273,7 @@ BOOL CALLBACK khrIcdOsVendorsEnumerate(PINIT_ONCE InitOnce, PVOID Parameter, PVO
     }
 
     // Add adapters according to DXGI's preference order
-    HMODULE hDXGI = LoadLibrary("dxgi.dll");
+    HMODULE hDXGI = LoadLibraryA("dxgi.dll");
     if (hDXGI)
     {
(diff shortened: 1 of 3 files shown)


jenatali (approved):

bashbaug:
Merging as discussed in the April 11th teleconference.

PR #217, opened 2023-04-11
Fix memory leaks related to layers and vendor icds

### LINKED ISSUE #157: Memory leaks in libOpenCL.so.1!khrIcdVendorAdd
Intel Inspector detects memory leaks for two allocations in khrIcdVendorAdd():

1. Error: Memory leak: 16 Bytes:
  >libOpenCL.so.1!khrIcdVendorAdd - /OpenCL-ICD-Loader/loader/icd.c:135
   libOpenCL.so.1!khrIcdOsVendorsEnumerate - /OpenCL-ICD-Loader/loader/linux/icd_linux.c:125

2. Error: Memory leak: 80 Bytes
  >libOpenCL.so.1!khrIcdVendorAdd - /OpenCL-ICD-Loader/loader/icd.c:153
   libOpenCL.so.1!khrIcdOsVendorsEnumerate - /OpenCL-ICD-Loader/loader/linux/icd_linux.c:125

Looking at the code in[ icd.c](https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/master/loader/icd.c) is seems that vendor structure is initialized and added into `khrIcdVendors` global list, but there is no explicit deinitialization function that would free those elements. Probably one of the options to fix that issue would be to add a function with `_attribute_((destructor))` on Linux or `DllMain()` handler on Windows to handle library unload event and deallocate those blocks.

JablonskiMateusz:
fixes: https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/157

### DIFF
--- a/loader/icd.c
+++ b/loader/icd.c
@@ -210,6 +210,15 @@ Done:
     }
 }
 
+void khrIcdVendorCleanup() {
+    KHRicdVendor* nextVendor = NULL;
+    while (khrIcdVendors) {
+        nextVendor = khrIcdVendors->next;
+        free(khrIcdVendors);
+        khrIcdVendors = nextVendor;
+    }
+}
+
 #if defined(CL_ENABLE_LAYERS)
 void khrIcdLayerAdd(const char *libraryName)
 {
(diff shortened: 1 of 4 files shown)


bashbaug:
Hi, I think this is no longer needed now after merging https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/268.

Can you please check?  Thanks!

PR #221, opened 2023-05-16
CTS cache

Kerilk:
This implements my suggestion from today. This may not fully resolve the issue but alleviate it greatly. It ensures some external dependencies are cached before the jobs are started. If this is an improvement, we could generalize to other platforms. 

There also exist ways to cache deb packages on Ubuntu to avoid having to re-download them each times:
```
      - uses: awalsh128/cache-apt-pkgs-action@latest
        with:
          packages: 
          version: 1.0
```
We could think about adopting these as well.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,6 +6,28 @@ env:
   OPENCL_PKGCONFIG_PATHS: ${{ github.workspace }}/install/lib/pkgconfig:${{ github.workspace }}/external/OpenCL-Headers/install/share/pkgconfig
 
 jobs:
+
+  tools:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        CMAKE: [3.21.2]
+    env:
+      CMAKE_URL: https://github.com/Kitware/CMake/releases/download/v${{ matrix.CMAKE }}/cmake-${{ matrix.CMAKE }}-Linux-x86_64.tar.gz
+
+    steps:
+    - name: Cache CMake
+      uses: actions/cache@v3
+      id: cmake
+      env:
+        cache-name: cache-cmake
+      with:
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
Nice, I'm sold, this looks like a good solution.

It'd be great to apply similar methodology to Linux or MacOS also, though that can be done in a separate commit.

My only minor comment is to perhaps reconsider the PR title since it will become the commit message?  It's not immediately clear to me what "CTS" stands for.

Thanks for the fast fix!  This should hopefully make CI both faster and more robust.

Kerilk:
I'll implement for the other platforms.

Note that in previous experiments I tried caching the tools folder, but the performance impact was significant. Anticipating questions here.

MathiasMagnus:
I also support this.

Once we start cracking on CI, we'll be sure to get all of this as well. I've been [testing the waters](https://github.com/MathiasMagnus/Test-FindOpenCL/blob/ci/.github/workflows/Consumption.yaml) with a unified script for all plats using this same caching mechanism. It has been working fine, I'm just stuck on a MacOS issue with XCode. I also cache a custom build of CMake that get rebuilt when the commit on the tip of the working branch changes. (It's testing repo for updating FindOpenCL.cmake, testing all manners of supported builds and consumption mechanisms.)

Kerilk:
OK, took me way too long to get it right, but I've added the cache for Linux as well. The Linux minimum uses the docker image which embeds CMake already, so the download is not required there.

We rely on brew for installing cmake and ninja on MacOS so if we wanted to cache those we should look at solutions like this one:
https://stackoverflow.com/a/65056232

But I would advise going at it incrementally, so unless I missed something, this one should be good to go.

MathiasMagnus (approved):
Should be good.

PR #227, opened 2023-11-20
cl_khr_icd 2.0.0 loader managed dispatch

Kerilk:
This implements the loader managed dispatching strategy discussed in https://github.com/KhronosGroup/OpenCL-Docs/issues/1003


loader/icd.c
```diff
@@ -132,27 +144,77 @@ void khrIcdVendorAdd(const char *libraryName)
         char *suffix;
         size_t suffixSize;
 
-        // call clGetPlatformInfo on the returned platform to get the suffix
-        if (!platforms[i])
+        // skip NULL platforms and non dispatchable platforms
+        if (!platforms[i] || !platforms[i]->dispatch)
```
s-barannikov, line 148: How can dispatch be NULL?

include/cl_khr_icd2.h
```diff
@@ -0,0 +1,80 @@
...
+#define CL_LAYER_API_VERSION_200 200
+
+typedef cl_int CL_API_CALL
+clDeinitLayer_t(void);
+
+typedef clDeinitLayer_t *pfn_clDeinitLayer;
+#endif //defined(CL_ENABLE_LAYERS) && !defined(CL_LAYER_API_VERSION_200)
+
+#if !defined(CL_ICD2_TAG_KHR)
+#define CL_ICD2_TAG_KHR ((size_t)0x4F50454E434C3331ULL)
```
s-barannikov, line 15: This doesn't fit size_t on 32-bit platforms, potentially causing compiler warnings.

Kerilk: Indeed the expected tag on 32 bit platforms would be `0x434C3331`. I will update the PR with a detection for 32 bit platforms.

loader/icd.c
Kerilk: I just extended the original check, to ensure that the returned platform is valid. The idea is not to register platform that obviously do not match the OpenCL icd constraints.

include/cl_khr_icd2.h
Kerilk: Found the warning on msvc:
`warning C4310: cast truncates constant value`

Kerilk: Replaced with:
```C
#if !defined(CL_ICD2_TAG_KHR)
#if INTPTR_MAX == INT32_MAX
#define CL_ICD2_TAG_KHR ((intptr_t)0x434C3331)
#else
#define CL_ICD2_TAG_KHR ((intptr_t)0x4F50454E434C3331)
#endif
```
Hopefully portable enough.

PR #230, opened 2024-01-31
Add OpenCL 3.0 API clGetKernelSuggestedLocalWorkSizeKHR.

haonanya:


Kerilk:
Thanks for suggesting this change. If we were to consider adding this to the dispatch table, it would first need to be added here:
https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_icd.h

Also the print layer code would have to be generated and updated as well.

Kerilk:
To be completely honest with you, I don't see this change moving further unless https://github.com/KhronosGroup/OpenCL-Docs/issues/1003 is adopted and we do a major release.

bashbaug (changes requested):
I don't think we should be doing this.  Frankly, I think it was a mistake to put any extension APIs into the OpenCL ICD loader.  It is too late to change this for the extension APIs that are already in the ICD loader, because we want to stay compatible, but we shouldn't be doing this for any extension APIs going forward.

See: https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/229410f86a8c8c9e0f86f195409e5481a2bae067/loader/icd_dispatch.c#L70-L73

Instead, to call an extension API an application should query the extension function pointer using `clGetExtensionFunctionAddressForPlatform`.  This can be done "manually" in the application code itself, or it can be done via a helper library, like the one in the OpenCL SDK:

https://github.com/KhronosGroup/OpenCL-SDK/tree/main/lib/src/Extensions

haonanya:
Thanks @bashbaug and @Kerilk , I will close the PR and issue.

PR #231, opened 2024-02-05
Explicitly mark compareLayer as __cdecl

jenatali:
There's command-line switches which can change the default calling convention for unannotated functions. Visual Studio defaults to `/Gd` which is `__cdecl`, but Windows defaults to `/Gz` which is `__stdcall`. That causes an x86 build break, since `qsort` requires the input parameter to be `__cdecl`. Just add the explicit annotation so it works regardless of the default.

### DIFF
--- a/loader/windows/icd_windows.c
+++ b/loader/windows/icd_windows.c
@@ -109,7 +109,7 @@ static WinLayer* pWinLayerBegin;
 static WinLayer* pWinLayerEnd;
 static WinLayer* pWinLayerCapacity;
 
-static int compareLayer(const void *a, const void *b)
+static int __cdecl compareLayer(const void *a, const void *b)
 {
     return ((WinLayer *)a)->priority < ((WinLayer *)b)->priority ? -1 :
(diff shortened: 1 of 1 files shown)


Kerilk (approved):
Nice one, thanks for fixing this! Once CI succeeds, we should be able to merge.

jenatali:
Looks like the CI run hit an infrastructure issue (timeout retrieving GPG key). I don't see an option to re-run it but I'm not super familiar with the GitHub actions UI. Might just be a permission issue though.

Kerilk:
Launched a rerun of the failed jobs.

Kerilk:
@bashbaug @MathiasMagnus Do we want to add the above configuration to the CI? Or merge this as is and try to remember about this?

MathiasMagnus:
This seems like a good change. We should probably test both calling conventions, but I'd defer it via an issue of its own.

MathiasMagnus (approved):

bashbaug:
I added an issue to track adding this case to CI, so we don't lose track of it.  Let's just merge this now.

PR #233, opened 2024-02-07
Fix memory leaks in loader_test.

Kerilk:
I was playing with valgrind and found a couple of memory leaks in the loader test, this patch addresses these.

### DIFF
--- a/test/driver_stub/cl.c
+++ b/test/driver_stub/cl.c
@@ -1835,11 +1835,13 @@ clEnqueueNativeKernel(cl_command_queue   command_queue ,
     return return_value;
 }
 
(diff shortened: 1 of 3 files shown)


bashbaug (approved):
Thanks for the fixes!

I'm a little surprised some of the other test_create_calls() do not require do not require releases also, but this is a definite improvement.

Kerilk:
> I'm a little surprised some of the other test_create_calls() do not require do not require releases also, but this is a definite improvement.

Yes, these are done in the `test_release`  function at the end, that puzzled me too. There are no leaks that valgrind would report that I know of anymore.

PR #234, opened 2024-02-15
Fix still reachable leaks in loader tests.

Kerilk:
This fixes additional leaks that were found using `valgrind`'s `--show-leak-kinds=all`.
The remaining leaks are in the stub driver itself and will be addressed in https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/227


bashbaug (approved):
Thank you for fixing these!

PR #250, opened 2025-04-01
OpenCL ICD Loader Deinitialization Support

Kerilk:
This is the follow up of https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/238 and the implementation of https://github.com/KhronosGroup/OpenCL-Docs/issues/1003.

It depends on https://github.com/KhronosGroup/OpenCL-Docs/pull/1350 and https://github.com/KhronosGroup/OpenCL-Headers/pull/278.

It implements OpenCL ICD Loader de-initialization.


Kerilk:
Will rebase.

loader/icd.c
```diff
@@ -472,3 +489,43 @@ void khrIcdContextPropertiesGetPlatform(const cl_context_properties *properties,
...
+    KHR_ICD_TRACE("Finalizing and unloading layers\n");
+    struct KHRLayer *head = khrFirstLayer;
+    deinitLayer.dispatch = khrDeinitDispatch;
+    khrFirstLayer = &deinitLayer;
+
+    while(head) {
+        struct KHRLayer *cur = head;
+#ifdef CL_LAYER_INFO
+	free(cur->libraryName);
+#endif
```
bashbaug: Indentation nitpick:

```suggestion
#ifdef CL_LAYER_INFO
	    free(cur->libraryName);
#endif
```

Kerilk:
> Thanks for continuing to work on this!
> 
> A few other high-level comments:
> 
>     * We'll eventually need to update the XML file with these changes, especially the new `CL_PLATFORM_UNLOADABLE_KHR` enum.  Organizationally, would we consider this an update to `cl_khr_icd`?
> 
>     * We should rev the loader version as part of these changes, see: `OPENCL_ICD_LOADER_VERSION_REV`.

Thanks for reviewing this!
Yes, this is the intent, and the PRs linked into the description should implement these changes.
I do consider it an update on cl_khr_icd but it applies to both 1.0 and 2.0 implementations, so I don't know how this works in practice. Just advertising support is enough to enable it, irrespective of the loader version used.

Kerilk:
> If we bump the project version in (2), can layers use this to figure out whether they can safely return version 200? I was considering updating my layers to version 200, but if I do this unconditionally it means that older ICD loaders will refuse to load them.

That is one of the things I wanted to brainstorm about. There are ways we could keep all layers version 100 of the API, use a new property that would advertise if the layer can be unloaded. We could then use a new initialization function (InitWithProperties maybe to address future problem like this one). If the layer is initialized with the old function, it would need to do cleanup (if required) in an at_exit callback, otherwise, the loader supports de-initialization of the layer, and will call the appropriate function if found. That is quite different from the current proposal, but could be a better, more future proof approach.

include/cl_khr_icd2.h
```diff
@@ -18,6 +18,19 @@
 
 #include <CL/cl.h>
 
+#if !defined(CL_PLATFORM_UNLOADABLE)
+#define CL_PLATFORM_UNLOADABLE_KHR 0x0921
```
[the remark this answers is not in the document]
Kerilk: Good catch thanks!

README.md
```diff
@@ -145,3 +145,5 @@ The following debug environment variables are available for use with the OpenCL
 | OPENCL_LAYERS                     | Specifies a list of layers to load. |  `export OPENCL_LAYERS=libLayerA.so:libLayerB.so`<br/><br/>`set OPENCL_LAYERS=libLayerA.dll;libLayerB.dll` |
 | OPENCL_LAYER_PATH                 | On Linux and Android, specifies a directory to scan for layers to enumerate in place of the default `/etc/OpenCL/layers'. | `export OPENCL_LAYER_PATH=/my/local/layers/search/path` |
 | OCL_ICD_ENABLE_TRACE              | Enable the trace mechanism          |  `export OCL_ICD_ENABLE_TRACE=True`<br/><br/>`set OCL_ICD_ENABLE_TRACE=True`<br/>`true, T, 1 can also be used here.`  |
+| OCL_ICD_FORCE_LEGACY_TERMINATION  | Force using the legacy termination scheme. This enables supporting outdated layers, but is problematic is libOpenCL is unloaded before the application exits |  `export OCL_ICD_FORCE_LEGACY_TERMINATION=True`<br/><br/>`set OCL_ICD_FORCE_LEGACY_TERMINATION=True`<br/>`true, T, 1 can also be used here.`  |
```
bashbaug: How about:

```suggestion
| OCL_ICD_FORCE_LEGACY_TERMINATION  | Forces using the legacy termination scheme. Legacy termination supports older layers, but does not support layer de-initialization and does not close library handles for layers or ICDs.  |  `export OCL_ICD_FORCE_LEGACY_TERMINATION=True`<br/><br/>`set OCL_ICD_FORCE_LEGACY_TERMINATION=True`<br/>`true, T, 1 can also be used here.`  |
```

loader/icd.c
```diff
@@ -472,3 +528,50 @@ void khrIcdContextPropertiesGetPlatform(const cl_context_properties *properties,
...
+        KHR_ICD_TRACE("ICD Loader deinitialization disabled\n");
+        return;
+    }
+
+    KHR_ICD_TRACE("ICD Loader deinitialization\n");
+
+#if defined(CL_ENABLE_LAYERS)
+    // free layers first in reverse order of their creation (front to back)
+    // they may still need to use vendors while terminating
+    KHR_ICD_TRACE("Finalizing and unloading layers\n");
```
Kerilk:
```suggestion
    KHR_ICD_TRACE("finalizing and unloading layers\n");
```

Kerilk:
Fixed a bug where when loading and unloading the OpenCL ICD Loader, environment variable values would get overwritten.

PR #253, opened 2025-06-04
Hide ICD-specific symbols from users

Kerilk:
This addresses some aspects of https://github.com/KhronosGroup/OpenCL-Docs/issues/1378
`clGetExtensionFunctionForPlatform` will always return `NULL` if called with `function_name` set as one of the functions defined in the `cl_khr_icd` extension.
`clGetExtensionFunction` would not return icd specific symbols (unless the platform extension suffix is "KHR") so I don't think we need to do anything there, bu I  could be persuaded otherwise.

### DIFF
--- a/loader/icd_dispatch.c
+++ b/loader/icd_dispatch.c
@@ -303,9 +303,25 @@ static inline void* clGetExtensionFunctionAddressForPlatform_body(
     // to get the extension function address.
 
     KHR_ICD_VALIDATE_HANDLE_RETURN_ERROR(platform, NULL);
+
+#define KHR_ICD_REJECT_EXTENSION_FUNCTION(name)                                \
+    do                                                                         \
+    {                                                                          \
+        if (!strcmp(function_name, #name))                                     \
+        {                                                                      \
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
I think this is fine.  In theory, an intercept layer could want to intercept calls to these functions, and with this change it would need to jump through some extra hoops to do so.  I've never had a need to do this with the OpenCL Intercept Layer, though, so I think the odds of this happening are small.

Noting for completeness, there are still mechanisms to get these functions, if needed for a sufficiently stubborn application, but it'd be a bit more work now.

Kerilk:
This will go through the layers, but the return from the platform will always be null (or I did not code the behavior I wanted).
Inside the loader we always use the raw pointer obtained from the library, so layers would not be invoked, but if as intercept library was used as a driver then it would provide it's own version.

How would you gain access to these from an application if these can be obtained neither through `clGetExtensionFunctionForPlatform` or `clGetExtensionFunction`? You would need to load the driver's yourself and get the function pointer from `clGetExtensionFunction`, correct?

PR #254, opened 2025-06-17
Fix nameless union initialization.

Kerilk:
This should address the issue generated by https://github.com/KhronosGroup/OpenCL-Headers/pull/280.
If we want backward compatibility, we should do some version detection to address both cases.

Error was, for reference:
```
/__w/OpenCL-ICD-Loader/OpenCL-ICD-Loader/loader/icd_dispatch_generated.c:6828:5: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
    &clGetPlatformIDs_disp,
    ^~~~~~~~~~~~~~~~~~~~~~
    {                     }
```


bashbaug:
Merging - this is needed to build with the latest headers.

Issue #256, opened 2025-07-25
only use brace initialization for clGetPlatformIDs when anonymous unions are supported


bashbaug:
We hit a problem recently in the OpenCL-CLHPP CI where OpenCL ICD loader builds with `CMAKE_C_EXTENSIONS=OFF` produced anonymous union / brace initialization warnings.  Because the CI builds with warnings-as-errors, this caused CI to fail.

https://github.com/KhronosGroup/OpenCL-CLHPP/pull/330

```
/home/bashbaug/git/OpenCL-ICD-Loader/loader/icd_dispatch_generated.c:6828:5: warning: braces around scalar initializer
 6828 |     {&clGetPlatformIDs_disp},
```

At least on my system, the difference between `CMAKE_C_EXTENSIONS=OFF` and the default `CMAKE_C_EXTENSIONS=ON` is that compilation with extensions off passes `-std=c99`, whereas compilation with extensions on is done with `-std=gnu99`.

If we want to build with extensions off and hence with `-std=c99` we should only use brace initialization when anonymous unions are supported.  This may be as simple as something like:

```diff
struct _cl_icd_dispatch khrMasterDispatch = {
+#if __CL_HAS_ANON_STRUCT__
    {
+#endif
    &clGetPlatformIDs_disp
+#if __CL_HAS_ANON_STRUCT__
    }
+#endif
    ,
...
```

Kerilk:
I think the problem is addressed, should we close this one?

PR #257, opened 2025-07-28
Address anonymous union initialization when extensions are disabled.

Kerilk:
This should address https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/256
This adds a macro to condition brace initialization to the presence of `__CL_HAS_ANON_STRUCT__`.


bashbaug (approved):
Thanks!  LGTM

bashbaug:
Merging as discussed in the August 5th teleconference.

PR #258, opened 2025-07-29
CMake C extensions test

Kerilk:
For discussion, since many dependencies were tested with C extensions off.


bashbaug:
Discussed in the August 5th teleconference.  OK to merge.

PR #260, opened 2025-08-17
Make android paths easier to override

rjodinchr:
At the moment, the OpenCL-ICD-Loader checkout in android is using a patch to override the vendor/layer paths staticaly in the source code.

This patch aims at dropping that patch in android. It also moves the overload to the build system file (Soong's blueprint for android) instead of having it staticaly inline in the source code.

https://android.googlesource.com/platform/external/OpenCL-ICD-Loader/+/refs/heads/main/loader/icd_platform.h


rjodinchr:
cc @Jeremy-Kemp

Kerilk (approved):
LGTM.

PR #263, opened 2025-09-04
Add static link libs to OpenCL.pc

dg0yt:
This is a (variation of) a patch added to the vcpkg port in order to fix downstream tests of ggml/whisper.cpp/llama.cpp with static library linkage.
- Static link libs are collected in `OPENCL_LIBS_PRIVATE_PC` and exported in the `Libs.private` field. This field is used by `pkg-config --static`.
- The generation of the pc file now takes place after configuring the OpenCL target.
- For `CMAKE_DL_LIBS`, only `dl` and the emtpy value are handled.


bashbaug:
@MathiasMagnus could you please have a look at this?

bashbaug:
Sorry for the delay, not sure why this wasn't merged previously but I don't see any reason to wait further.

PR #264, opened 2025-10-03
Switch to __CL_HAS_ANON_UNION__ macro

joselopeqti:
Update loader to use __CL_HAS_ANON_UNION__ macro introduced in https://github.com/KhronosGroup/OpenCL-Headers/pull/293


Kerilk (approved):
To be merged if https://github.com/KhronosGroup/OpenCL-Headers/pull/293 ends up being merged. Rerun CI on it is.

PR #266, opened 2025-10-22
Dispatch table generation.

Kerilk:
This factors the dispatch table code that was duplicated between the test layer and the loader.
It also renames khrMasterDispatch to khrMainDispatch.


Kerilk:
None that I can think of.

PR #268, opened 2026-01-22
Added an OpenCL deinitialization test and plugin infrastructure.

Kerilk:
Hello,

This PR does several things:

- Refactor platform specific components (and dependencies) of the loader in logical units;
- Creates a test program (cl_plugin_loader_test) that loads and unloads a shared library plugin in a loop (leverages above components);
- Creates a plugin, linked to the OpenCL ICD Loader that will exercise OpenCL calls when initialized through a predetermined entry point. When unloaded, this plugin will also automatically unload the OpenCL ICD Loader.
- Adds a new test that will invoke the cl_plugin_loader_test program using the plugin, exercising the OpenCL ICD Loader loading and unload capabilities.

For now the test just verifies that the program returns correctly, but additional manual testing can be done using loader tracing and memory validator like `valgrind` to ensure proper execution and absence of memory leaks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,43 +230,6 @@ endif()
 if((CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME OR OPENCL_ICD_LOADER_BUILD_TESTING) AND BUILD_TESTING)
     add_subdirectory (test)
 
-    set (CL_PLUGIN_LOADER_TEST_SOURCES
-        test/plugin_test/cl_plugin_loader_test.c
-        test/plugin_test/cl_plugin.h
-        loader/icd_envvars.h
-        loader/icd_trace.c
-        loader/icd_trace.h
-        loader/icd_library.h
-    )
-
-    if (WIN32)
-        list (APPEND CL_PLUGIN_LOADER_TEST_SOURCES
-            loader/windows/icd_windows_formats.h
-            loader/windows/icd_windows_envvars.c
-            loader/windows/icd_windows_library.c)
-        # Only add the DXSDK include directory if the environment variable is
-        # defined.  Since the DXSDK has merged into the Windows SDK, this is
-        # only required in rare cases.
-        if (DEFINED ENV{DXSDK_DIR} AND NOT (MINGW OR MSYS OR CYGWIN))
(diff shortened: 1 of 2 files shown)


loader/linux/icd_linux_library.c
```diff
@@ -0,0 +1,51 @@
...
+/*
+ *
+ * Dynamic library loading functions
+ *
+ */
+
+// dynamically load a library.  returns NULL on failure
+void *khrIcdOsLibraryLoad(const char *libraryName)
+{
+    void* ret = dlopen (libraryName, RTLD_NOW);
```
EwanC: I know this inconsistency was here before but since we're touching the code
```suggestion
    void* ret = dlopen(libraryName, RTLD_NOW);
```

loader/windows/icd_windows_library.c
```diff
@@ -0,0 +1,61 @@
...
+    return GetProcAddress( (HMODULE)library, functionName);
+}
+
+// unload a library.
+void khrIcdOsLibraryUnload(void *library)
+{
+    FreeLibrary( (HMODULE)library);
+}
+
+
```
EwanC: nitpick, could delete these trailing lines

Kerilk:
Thanks for the great review! I"ll go an fix these ASAP.

CMakeLists.txt
```diff
@@ -223,7 +229,46 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME OR OPENCL_ICD_LOADER_BUILD_TESTING)
 endif()
 if((CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME OR OPENCL_ICD_LOADER_BUILD_TESTING) AND BUILD_TESTING)
     add_subdirectory (test)
-endif()
+
+    set (CL_PLUGIN_LOADER_TEST_SOURCES
```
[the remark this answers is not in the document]
Kerilk: Used the first strategy, seems to be working.
We can revisit later by using something like this:
https://github.com/Kerilk/OpenCL-ICD-Loader/commit/ad3553091f6c7761aae636839ffb47ce01a5eeed

PR #271, opened 2026-02-09
Add cl_khr_icd instance support.

Kerilk:
This implements https://github.com/KhronosGroup/OpenCL-Docs/pull/1527

### DIFF
--- a/loader/icd.c
+++ b/loader/icd.c
@@ -76,6 +76,8 @@ void khrIcdVendorAdd(const char *libraryName)
 #if defined(CL_ENABLE_LOADER_MANAGED_DISPATCH)
     clIcdGetFunctionAddressForPlatformKHR_fn p_clIcdGetFunctionAddressForPlatform = NULL;
     clIcdSetPlatformDispatchDataKHR_fn p_clIcdSetPlatformDispatchData = NULL;
+    clIcdCreateInstancePlatformKHR_fn p_clIcdCreateInstancePlatform = NULL;
+    clIcdDestroyInstancePlatformKHR_fn p_clIcdDestroyInstancePlatform = NULL;
 #endif
     cl_uint i = 0;
     cl_uint platformCount = 0;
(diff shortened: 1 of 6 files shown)


test/driver_stub/cl.c
```diff
@@ -19,6 +19,12 @@
 #include "icd_structs.h"
 #include "cl_khr_icd2.h"
 
+#if defined(CL_ENABLE_ICD2)
+#include "cl_khr_icd2.h"
```
EwanC: we're already unconditionally including this header a couple of lines above

loader/icd.c
```diff
@@ -253,53 +275,47 @@ void khrIcdVendorAdd(const char *libraryName)
             if (CL_SUCCESS != result)
             {
                 KHR_ICD_TRACE("found cl_khr_icd_unloadable but clGetPlatformInfo CL_PLATFORM_UNLOADABLE_KHR query failed\n");
-                free(vendor);
-                continue;
+                goto Error;
             }
         }
+        free(extensions);
```
EwanC: I might be misreading the diff, but we're also doing `free(extensions)` at the Error label, is it possible to double free if we do a `goto Error` after this?

loader/icd.c
```diff
@@ -315,7 +331,17 @@ void khrIcdVendorAdd(const char *libraryName)
         lastVendor = vendor;
 
         KHR_ICD_TRACE("successfully added vendor %s with suffix %s\n", libraryName, suffix);
-
+        continue;
+Error:
+        free(extensions);
+        free(suffix);
+#if defined(CL_ENABLE_LOADER_MANAGED_DISPATCH)
+        if (instance_platform)
```
EwanC, line 339: We can `goto Error` if the `vendor` malloc fails and returns nullptr, so don't know if this should be
```cpp
if (instance_platform && vendor)
```

EwanC (commented):
I don't have much  of an insight on if this is the best solution, so my comments are more general observations.

loader/icd.c
Kerilk: Awesome catch, thanks, I knew there must have been some logic error left somewhere. One day we'll be able to use `defer`.

loader/icd.c
Kerilk: You're correct, I'll rely on the loop variable rather than the stored pointer.

test/driver_stub/cl.c
Kerilk: Will fix!

Kerilk:
> I don't have much of an insight on if this is the best solution, so my comments are more general observations.

Thanks a lot for the review!

PR #272, opened 2026-03-30
feat: add support for GNU/Hurd

pinotree:
Nothing else is needed beside using the bits of other Unix OSes.


bashbaug (approved):
Thank you for your contribution!

Note, we'll try to keep this working on a best-effort basis, but without any sort of CI coverage we can't make any promises.  If you find any issues in the future though, please don't hesitate to let us know!

PR #275, opened 2026-05-07
Prevent windows library unloading at process termination.

Kerilk:
DLL unloading at process de-initialization should be avoided:

> Similarly, the entry-point function must not call the [FreeLibrary](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-freelibrary) function (or a function that calls FreeLibrary) during process termination, because this can result in a DLL being used after the system has executed its termination code.

https://learn.microsoft.com/en-us/windows/win32/dlls/dllmain

### DIFF
--- a/loader/cllayerinfo.c
+++ b/loader/cllayerinfo.c
@@ -121,6 +121,15 @@ static void run_silently(void (*pfn)(void))
     atexit(silence_layers);
 }
 
+static void run_silently_int(void (*pfn)(int), int arg)
+{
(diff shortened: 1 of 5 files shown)


jenatali (approved):

bashbaug (approved):

joselopeqti (approved):

bashbaug:
Merging as discussed in the May 19th teleconference.

PR #273, opened 2026-05-05
OpenCL ICD Loader updates for OpenCL 3.1

bashbaug:
This PR contains ICD loader updates for OpenCL 3.1.

PR #277, opened 2026-05-13
update one more ICD loader version for OpenCL 3.1

bashbaug:
When I made the OpenCL ICD loader updates for OpenCL 3.1 in #273 I missed one spot that needed updating.


Kerilk (approved):
Good catch.

PR #278, opened 2026-05-14
remove use of the cl_khr_icd2.h header

bashbaug:
The cl_khr_icd2.h header should not be needed anymore since all of the necessary support is in the upstream headers.

This also fixes a problem with redefinition warnings for some compilers.


Kerilk (approved):
Great, shows we have correctly upstreamed everything.

PR #280, opened 2026-06-05
Remove use of the cl_khr_icd2.h header

joselopeqti:
Removed header but not usage in another file.


bashbaug (commented):
Can we remove the cl_khr_icd2.h header file from this repo completely now? I think this was the last use.

I don't remember why I didn't remove it previously as part of #278, weird.

Kerilk:
Because I need it there: https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/271

PR #281, opened 2026-06-19
Use clGetLayerInfo api_version to select layer init path

joselopeqti:
Currently the loader accepts only CL_LAYER_API_VERSION_100 layers and uses a single global flag (OCL_ICD_FORCE_LEGACY_TERMINATION) to switch between clInitLayer (v1.0) and clInitLayerWithProperties (v2.0). This
broke backward compatibility: layers built against ICD loader v1.0 could not be loaded by the v2.0 loader without an environment-variable override, which is unavailable in sandboxed environments such as Android apps.

This change uses the api_version returned by clGetLayerInfo(CL_LAYER_API_VERSION) to automatically select the correct initialization path per layer, without any user intervention:

  CL_LAYER_API_VERSION_100 (100) — v1.0 layer
    Uses clInitLayer. No clDeinitLayer is expected; the layer manages its own teardown via atexit().

  CL_LAYER_API_VERSION_200 (200) — v2.0 layer  [new constant]
    Uses clInitLayerWithProperties. clDeinitLayer is optional.

  Any other value — rejected with a descriptive trace message.

The dispatch call block now branches on which init function pointer was resolved (p_clInitLayer vs p_clInitLayerWithProperties) rather than on the global khrForceLegacyTermination flag.

khrForceLegacyTermination (OCL_ICD_FORCE_LEGACY_TERMINATION) is preserved as a last-resort escape hatch: when set it overrides the version-based path and treats all layers as v1.0, for exceptional cases where the layer's clGetLayerInfo reports an incorrect version.

### DIFF
--- a/loader/icd.c
+++ b/loader/icd.c
@@ -22,6 +22,22 @@
 #include <stdlib.h>
 #include <string.h>
 
+/*
+ * CL_LAYER_API_VERSION_200 identifies layers built against ICD loader v2.0.
+ * Such layers export clInitLayerWithProperties instead of clInitLayer, and
+ * optionally export clDeinitLayer for explicit teardown.
+ *
+ * CL_LAYER_API_VERSION_100 (defined in <CL/cl_layer.h>) identifies layers
+ * built against ICD loader v1.0 that export only clInitLayer.
+ *
+ * The loader selects the initialization path based on the version returned by
+ * clGetLayerInfo(CL_LAYER_API_VERSION), ensuring both generations of layers
+ * are supported without requiring any environment-variable workaround.
+ */
+#ifndef CL_LAYER_API_VERSION_200
+#define CL_LAYER_API_VERSION_200 200
+#endif
+
 KHRicdVendor *khrIcdVendors = NULL;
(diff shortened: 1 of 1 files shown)


Kerilk:
The reason we discarded this approach is that the timing of `at_exit` and the timing of the loader termination are unrelated, and the race makes the two termination mechanisms incompatible.

The forcing mechanism allows new layers to use the old at_exit behavior. Notice that this behavior is incompatible with using the OpenCL ICD loader inside a plugin that would get loaded and unloaded, and leaks memory, so in general layers that use at exit are just unsafe to use in the general case (note that the loader is unsafe to use in legacy termination mode as well irrespective of layers). That is why the mechanism to enable legacy termination is opt-in, and not a detection that at least one layer is using `at_exit`.

Maybe we should message better that layers that use `at_exit` are deprecated, my original implementation would just reject layers that use `at_exit`, but we wanted to have an escape hatch for cases like yours.

In your case we could devise a configuration bit that the loader could read if environment variables cannot be used, probably in the configuration files.

PR #288, opened 2026-09-16
Skip disabled devices when enumerating ICDs via HKR

manocharahul:
### Problem

On Windows, `ProbeDevice()` in the HKR enumeration path only screens for reboot-pending states. A device that has been **disabled** in Device Manager is still reported as `Valid`.

Disabling a device does not remove its registry values, so the loader goes on to read `OpenCLDriverName` from the disabled adapter's HKR software key and registers an ICD for hardware that can never supply an OpenCL device.

The result is a **platform with zero devices**. That is legal, and `clinfo` handles it, but it is not useful — and applications that treat `CL_DEVICE_NOT_FOUND` as fatal fail to start even when another GPU in the system is present and working.

### Why the HKR path specifically

Enumeration uses `CM_GETIDLIST_FILTER_CLASS | CM_GETIDLIST_FILTER_PRESENT`. In Windows terms *present* means physically installed — a disabled device is still present — so it is returned by `CM_Get_Device_ID_ListW()` and `ProbeDevice()` is the only gate after that.

The DXGK path is unaffected: it uses `D3DKMTEnumAdapters2`, which returns started adapters only.

### Fix

Reject devices reporting `CM_PROB_DISABLED` so their ICD is never loaded. Both existing call sites already use `if (ProbeDevice(...) != Valid) continue;`, so no caller changes are needed, and `CM_PROB_DISABLED` comes from `cfgmgr32.h`, which is already included.

The check is deliberately narrow. It screens only the unambiguous, persistent case — a device a user explicitly disabled. `CM_PROB_FAILED_START` and a broader `!(ulStatus & DN_STARTED)` test were both considered and left out: the former can be transient (mid driver install, resource conflict), and `ProbeDevice()` is also called on software-component devnodes, where `DN_STARTED` semantics are less certain and a false rejection would hide a working ICD.

### Verification

Reproduced on a two-GPU Windows system with one GPU disabled in Device Manager:

- **Before:** two platforms, one reporting `0` devices. Loader trace shows the disabled device accepted and its ICD path read.
- **After:** one platform with the working GPU. Trace shows `WARNING: device is disabled (0x1802400), skipping...`, and the disabled device's `OpenCLDriverName` is never read or its ICD loaded.

`clinfo` exits `0` in both cases; the difference is that the useless empty platform is no longer advertised.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jenatali (approved):
Seems reasonable to me.

manocharahul:
I have signed CLA but its still not updating. Can someone please check.

bashbaug:
> I have signed CLA but its still not updating. Can someone please check.

You need to address this:



> **Rahul Manocha** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

manocharahul:
I added my email to my profile and commit shows verified user now. Signed the CLA again, but doesn't seem to get updated

manocharahul:
CLA is signed now. Can we get workflow approval?

manocharahul:
@bashbaug  How do i get the workflow approved?

bashbaug (approved):
Merging as discussed in the September 22nd teleconference.

